PR #16, opened 2023-12-14
Add wasm support via webtransport (xwt)

cBournhonesque:
# Adding WASM support

This PR:
- adds a webtransport-wasm transport client
- updates the `interest_management` example to work with wasm


Nul-led (approved):

MOZGIII:
I'll step in to check what's wrong with xwt datagrams (hopefully within a few days).

lightyear/src/transport/webtransport/client_wasm.rs
```diff
@@ -0,0 +1,176 @@
...
+impl Transport for WebTransportClientSocket {
+    fn local_addr(&self) -> SocketAddr {
+        self.client_addr
+    }
+
+    fn listen(self) -> (Box<dyn PacketSender>, Box<dyn PacketReceiver>) {
+        let client_addr = self.client_addr;
+        let server_addr = self.server_addr;
+        let (to_server_sender, mut to_server_receiver) = mpsc::unbounded_channel();
+        let (from_server_sender, from_server_receiver) = mpsc::unbounded_channel();
```
MOZGIII, line 59: This can exhaust all available memory unless there is some other way to limit the amount of in-flight data in place

lightyear/src/transport/webtransport/client_wasm.rs
```diff
@@ -0,0 +1,176 @@
...
+            .collect::<Array>(),
+    );
+}
+
+impl Transport for WebTransportClientSocket {
+    fn local_addr(&self) -> SocketAddr {
+        self.client_addr
+    }
+
+    fn listen(self) -> (Box<dyn PacketSender>, Box<dyn PacketReceiver>) {
```
MOZGIII, line 55: It is a bit confusing: why does `listen` connects to somewhere? Listening won't work in the browser obviously, so why is it implemented even?

lightyear/src/transport/webtransport/client_wasm.rs
MOZGIII: In fact, a channel with no buffer might be useful here; `tokio` doesn't have these at all, but `flume` with size `0` is essentially that.

lightyear/src/transport/webtransport/client_wasm.rs
cBournhonesque: It could be more clear I think; but `listen` is the function I use to create the listener/receiver io channels.
I'll think about a better name

lightyear/src/transport/webtransport/client_wasm.rs
cBournhonesque: I am adding a bandwidth cap in another PR, but I think that's a good callout; i'll add a TODO

PR #76, opened 2024-01-13
wip

cBournhonesque:


lightyear/src/netconnection/steam/client.rs
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    pub fn connect(&mut self) {
+        let options = Vec::new();
+        let connection = self
+            .client
+            .networking_sockets()
+            .connect_p2p(NetworkingIdentity::new_steam_id(self.steam_id), 0, options)
+            .unwrap();
+    }
```
tbillington, line 54: ideally this returns a result since it contains an unwrap

lightyear/src/netconnection/steam/client.rs
```diff
@@ -0,0 +1,208 @@
...
+        }
+
+        None
+    }
+
+    pub fn client_id(&self, steam_client: &steamworks::Client<ClientManager>) -> ClientId {
+        steam_client.user().steam_id().raw()
+    }
+
+    pub fn disconnect(&mut self) {
```
tbillington, line 121: could take an optional reason here

lightyear/src/netconnection/steam/server.rs
```diff
@@ -0,0 +1,186 @@
...
+    fn disconnect_all(&mut self, flush_last_packets: bool) {
+        self.connections.keys().for_each(|client_id| {
+            self.disconnect_client(*client_id, flush_last_packets);
+        });
+    }
+}
+
+impl NetServer for SteamNetServer {
+    /// Update server connections, and receive packets from the network.
+    fn try_update(&mut self, delta_ms: f64, io: &mut Io) -> Result<ConnectionEvents> {
```
tbillington, line 95: food for thought, maybe a later update

In order to avoid repeated unnecessary allocations this could return an iterator or reuse the `Vec`s in `ConnectionEvents`

lightyear/src/netconnection/server.rs
```diff
@@ -0,0 +1,26 @@
...
+
+/// Event emitted on each server update, to notify a new client connections or disconnections.
+pub struct ConnectionEvents {
+    pub connected: Vec<ClientId>,
+    pub disconnected: Vec<ClientId>,
+}
+
+pub trait NetServer {
+    /// Return the list of connected clients
+    fn connected_client_ids(&self) -> Vec<ClientId>;
```
tbillington, line 16: Preferably this wouldn't force an allocation if the user didn't need it, preferring to return a reference or iterator.

Unfortunately rust doesn't support `fn connected_client_ids(&self) -> impl Iterator<Item = ClientId>;` on a trait at the moment :(

An intermediate step would be returning [`Keys`](https://doc.rust-lang.org/std/collections/hash_map/struct.Keys.html) but it doesn't feel _great_

Not a huge thing, I doubt this will be called very often.

lightyear/src/netconnection/mod.rs
```diff
@@ -0,0 +1,12 @@
...
+pub(crate) mod server;
+
+pub(crate) mod steam;
+
+pub(crate) mod netcode;
+
+/// The client id from a connect token, must be unique for each client.
+///
+/// Note that this is not the same as the [`ClientId`], which is used by the server to identify clients.
+pub type ClientId = u64;
```
tbillington, line 12: Ideally if this could possibly name collide with another type it should be a newtype eg `pub struct ClientId(u64)` so the compiler will appropriate reject it if you use the wrong one.

tbillington (approved):
LGTM :) had some nitpicks but really no showstoppers

cBournhonesque:
> LGTM :) had some nitpicks but really no showstoppers

Thanks for the review! I don't think this is ready yet, I need to actually test it.
But basically the objective of the PRs are:
- currently connections are based on the [netcode](https://github.com/mas-bandwidth/netcode/blob/main/STANDARD.md) protocol, which establishes a connection using a unreliable unordered transport like UDP with the use of a `ConnectToken`
- we already have an struct `Io` that allow us to switch between various transports (UDP, WebTransport, Channels...)
- however there might be some cases where we want to use a different "Connection" abstraction. For example steam networking sockets handle establishing the connection (even p2p or with NAT traversal), so I want to introduce traits that would let us potentially integrate with those

PR #85, opened 2024-01-15
added xpbd_3d with simple example based on leafwing_input example

mcbernie:
### simple integration of xpbd_3d.

example copied from existing leafwing_input example with required changes from 2d to 3d world


cBournhonesque:
Super cool!
I've never done anything 3D with bevy before.

Some thoughts:
- the server doesn't start in `headless` mode (because `Assets` doesn't exist without the Render plugin loaded), do you think you could make it work in headless mode?
- the prediction of the other player is pretty buggy; maybe the interpolation for Rotation needs a `slerp` instead of a `lerp`? (just something I read online). It's also possible that the prediction of other players doesn't work too well at the moment, it's also a bit jittery in 2D. (although we can make it smooth just by adding back input delay if we want)
- I thought the view/controls were a bit unintuitive, it was a bit hard to figure out what was going on. Maybe it would be better to have a more top-down view of the box, and have WASD only move the box in the x/y axis? Or maybe have a way to move the camera position?

mcbernie:
Hi
I did some corrections and optimizations.
The interpolation of positions and rotations in 3d Space seems buggy. I keep work on it ;)

- headless mode works now. 
   The example always add Colliders, but only Meshes when RenderPlugin does exists. *bevy_xpbd_3d* uses Assets<Mesh> in some Systems, when feature async-collider is enabled.
   
-  better camera position and controlling.
    Position the camera more like Top Down and rearranged the controlling behavior. Feels better now ;)

- prediction of other players
   For now, the rotation interpolation uses Slerp. But it looks buggy sometimes. Mostly on heavy impact of both players. I try to figure out what's exactly happens.

PR #98, opened 2024-01-19
fix lightyear version for lightyear-benches

rbouar:
This merge fixes the following compilation error:
```
error: failed to select a version for the requirement `lightyear = "^0.6.0"`
candidate versions found which didn't match: 0.7.0
...
```


cBournhonesque:
thanks!

PR #102, opened 2024-01-20
Allow WebTransport clients to gracefully disconnect

Ant59:
WebTransport clients disconnecting from the server causes a flood of errors in the server's log. This is due to the "netcode client" being cleaned up on disconnect, but the server is still looping the WebTransport client handling logic.

This PR fixes this by adding an option to bail from the loop when the connection is closed.


cBournhonesque:
Indeed; thanks for the fix!

PR #103, opened 2024-01-21
Use runtime arguments in Divan benchmarks

nvzqz:
This greatly reduces compile times and is not limited to arrays/slices.


cBournhonesque:
Great change, thanks!

PR #155, opened 2024-02-19
Fix outstanding errors for Bevy 0.13

Ant59:
I've made an attempt to fix up the remaining compilation, linting and testing errors in Lightyear with the Bevy 0.13 update.

However, I think for the Leafwing Input Manager feature to compile, that crate needs to update.

### DIFF
--- a/lightyear/Cargo.toml
+++ b/lightyear/Cargo.toml
@@ -30,7 +30,7 @@ webtransport = [
   "dep:tokio",
   "dep:ring",
 ]
-leafwing = ["dep:leafwing-input-manager", "lightyear_macros/leafwing"]
+leafwing = ["dep:leafwing-input-manager", "lightyear_macros/leafwing", "bevy/bevy_asset"]
 xpbd_2d = ["dep:bevy_xpbd_2d"]
 websocket = [
   "dep:tokio",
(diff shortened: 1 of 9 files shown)


lightyear/Cargo.toml
```diff
@@ -30,7 +30,7 @@ webtransport = [
   "dep:tokio",
   "dep:ring",
 ]
-leafwing = ["dep:leafwing-input-manager", "lightyear_macros/leafwing"]
+leafwing = ["dep:leafwing-input-manager", "lightyear_macros/leafwing", "bevy/bevy_asset"]
```
cBournhonesque, line 33: is bevy asset necessary for leafwing?

Ant59: It was necessary for the tests to run in the Leafwing feature (for use bevy::asset::AsyncReadExt), but I think it's not necessary for the build. I'm not entirely sure how to convince Cargo to only include a dependency for test with a certain feature though.

cBournhonesque:
Thanks! I integrated most of your changes.
I don't wnat to change bitcode too much as that part of the code can be a bit fragile

Ant59:
No problem. The bitcode changes were just ported from the latest commit on the bitcode repo to fix issues in Rust nightly. Works fine with Rust stable until next version.

PR #156, opened 2024-02-21
Explain where derive macros are from

doonv:
While reading the tutorial, I was a bit confused where the  `Add` and `Mul` macros came from. So I added a comment explaining it. Along with an explaination for `Deref` and `DerefMut`.


Nul-led:
lgtm

Nul-led (approved):

cBournhonesque:
Great point, thanks!

PR #158, opened 2024-02-21
Add Steam connection

cBournhonesque:
This PR adds a new `NetConnection` implementation: Steam!
(lightyear distinguishes between `NetConnection`, which is a long-term connection opened between client and server, and `Transport`, which is the mechanism to send raw datagrams over the network. In the past all the `NetConnections` were using the netcode.io protocol over unreliable transports (webtransport datagrams, UDP, etc.). This Steam `NetConnection` handles **both** the connection abstraction and the sending/receiving of datagrams, it's not using the netcode.io protocol at all)

### Remaining todos
- maybe provide a separate SteamP2P (or it could just be part of the configuration) mode where we use [ISteamNetworkingSockets::CreateListenSocketP2P](https://partner.steamgames.com/doc/api/ISteamNetworkingSockets#CreateListenSocketP2P) and [ISteamNetworkingSockets::ConnectP2P](https://partner.steamgames.com/doc/api/ISteamNetworkingSockets#ConnectP2P) to use the Steam Datagram Relay service to transmit data (handles NAT traversal, etc.)

NOTE: this can also be used for dedicated servers! read more here: https://partner.steamgames.com/doc/features/multiplayer/steamdatagramrelay

Another advantage is that it protects against DDOS by not leaking the IP address; SDR works either by providing the steam id directly, or by providing fake IP addresses that only work on SDR

### How to run

Test steam connection by going to the simple_box example: `cd examples/simple_box`
Run server: `cargo run -- server`
Run client: `cargo run -- client`
(update the settings in `assets/settings.ron` to use steam)


lightyear/src/connection/steam/client.rs
```diff
@@ -0,0 +1,159 @@
...
+        matches!(
+            self.connection_state(),
+            Ok(NetworkingConnectionState::Connected)
+        )
+    }
+
+    fn try_update(&mut self, delta_ms: f64) -> Result<()> {
+        if let Ok(single) = self.single_client.0.read() {
+            // TODO: the rest of the function probably shouldn't run when we can't get the lock?
+            single.run_callbacks();
```
msvbg, line 99: This should probably not be automatically called from Lightyear. If the user uses other Steam features then they might end up calling this multiple times per frame

cBournhonesque: The docs say that this should be fine to be called once per render frame: https://partner.steamgames.com/doc/api/steam_api#SteamAPI_RunCallbacks

Or is the issue that `run_callbacks` does both:
- forward connection events for the networking_sockets api
- run user-defined callbacks (which the user might want to have control over when to run)?

cBournhonesque: I see the bevy_steamworks also runs automatically once per frame, so I might just keep it that way for now until I understand steam callbacks better: https://github.com/HouraiTeahouse/bevy_steamworks/blob/main/src/lib.rs#L246

msvbg: I mean that the user might already be running callbacks as part of e.g bevy_steamworks or by using some other features from Steam, so this would effectively run callbacks multiple times.

zwazel: I don't see why someone would be using `bevy_steamworks` when he uses `lightyear`?
But `lightyear` probably should expose the steam client so that the user can also access the steam api whenever he wants, without the need of creating his own client. but in a way that he has no access to the `SingleClient` maybe? So he can't run `.run_callbacks`?

cBournhonesque: I think for now the simplest is to just merge what I have now.
What I expose to clients currently is mostly `dyn NetClient` (https://github.com/cBournhonesque/lightyear/blob/8f3ba758080e5952f26c4b91f4d6a8a1a2422787/lightyear/src/connection/client.rs#L121) so there's no steam-specific information, but I can think about how to improve things in a future PR.

I'll also want to investigate the steam P2P apis in a different PR.

msvbg: > I don't see why someone would be using bevy_steamworks when he uses lightyear

@zwazel Steam offers a variety of features beyond networking, and as such it makes a lot of sense for the Steam client and its update schedule to be owned outside of any networking library.

PR #166, opened 2024-02-26
handle WebTransport client disconnection

molixianggu:
This pull request addresses the issue of removing clients upon disconnection in WebTransport. 
I've introduced a somewhat unconventional method that leverages std::io::Error to extract SocketAddr upon disconnection, enabling the closure of the connection.

This change is minor but not the optimal solution; I aim to provide a direction for other developers.

### DIFF
--- a/lightyear/src/connection/netcode/error.rs
+++ b/lightyear/src/connection/netcode/error.rs
@@ -1,3 +1,5 @@
+use std::net::SocketAddr;
+use std::fmt;
 use thiserror::Error;
 
 /// The result type for all the public methods that can return an error in this crate.
@@ -24,3 +26,16 @@ pub enum Error {
     #[error(transparent)]
     Io(#[from] std::io::Error),
 }
+
+#[derive(Debug)]
+pub struct DisConnectionError {
+    pub addr: SocketAddr,
+}
+
+impl fmt::Display for DisConnectionError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "Error with socket address: {:?}", self.addr)
+    }
+}
+
+impl std::error::Error for DisConnectionError {}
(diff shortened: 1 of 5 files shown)


cBournhonesque:
Thanks, what is the exact problem you are trying to solve?

1. that when a client gets disconnect, the server gets flooded with error logs. But normally the futures handling that connection should be closed? Normally this should be handled here: https://github.com/cBournhonesque/lightyear/pull/166/files#diff-58c9e1c94459de992415bd079b7ff64ad8f142106c0e7a87d91ff8e4c389f78aR113
When the connection gets closed, these tasks should get dropped automatically. But i've noticed that it doesn't seem to work reliably.

2. that when the transport times out (for instance webtransport), the netcode connection should stop trying to send/receive packets?

molixianggu:
Thank you for the explanation. 
I spent some time trying. 
call the `cancel().await` method on `from_client_handle` and `to_client_handle` after `connection.closed().await;` and the tasks were successfully cancelled, [this](https://github.com/molixianggu/lightyear/blob/d4f41bae284142181f0c7305815c1d93ff9f5ea4/lightyear/src/transport/webtransport/server.rs#L123). However, I'm not sure why they weren't automatically deleted. Perhaps actively calling cancel is a solution. 

Additionally, I agree with what you mentioned about handling these cases when the connection is closed. Therefore, I updated the code and encapsulated a disconnect exception to pass the address.

cBournhonesque:
> Thank you for the explanation. I spent some time trying. call the `cancel().await` method on `from_client_handle` and `to_client_handle` after `connection.closed().await;` and the tasks were successfully cancelled, [this](https://github.com/molixianggu/lightyear/blob/d4f41bae284142181f0c7305815c1d93ff9f5ea4/lightyear/src/transport/webtransport/server.rs#L123). However, I'm not sure why they weren't automatically deleted. Perhaps actively calling cancel is a solution.
> 
> Additionally, I agree with what you mentioned about handling these cases when the connection is closed. Therefore, I updated the code and encapsulated a disconnect exception to pass the address.

For 1., I tried adding the `cancel().await` but I think in some cases the task just keeps running.. Just dropping the handle is normally the best way to cancel the task.

By the way are you on the bevy discord? I'd love to chat with you about this! There is a #lightyear channel!

Issue #115, opened 2024-01-29
feat: Multi Transport Servers (Cross Platform Dev)


Nul-led:
# Notes
- Allows for clients with multiple backgrounds to connect to the same game server (eg. Web, Native and Steam) which enables easy cross platform development

PR #169, opened 2024-02-27
Enable using multiple concurrent ServerConnections 

cBournhonesque:
# Context

This PR allows the user to create a `Server` that uses multiple `ServerConnections` at the same time.

(for context, a `Transport` is a trait to transmit raw bytes (UDP, WebTransport, WebSockets, local channels),
 a `ServerConnection` is an abstraction over a stateful connection.
 We will have 2 main `ServerConnection`s:
 - the `Netcode` protocol over a given `Transport`
 - using the steam sockets )
 
 With this change, we can provide multiple different `NetConfig`s when creating the server, and clients can connect to the server using any of them.
 
 - this works even if two different ServerConnections share the same `ClientId` (for example a user has steam id 1, and another users connects via netcode with client-id 1)
 
This potentially enables:
- crossplay between users on different platforms: some users can be connected via steam, while others aren't
- "ListenServer" behaviour where a player acts as host: they would create a Server that uses a `ServerConnection` with local channels to network with themselves, and uses a separate `ServerConnection` to connect with other players.

# Remaining todos:
- add unit tests for multi-transport, and multi-transport with ID collision
- update docs
- update all examples to support multi-transport, and figure out how to update clap to let users easily do it

# Potential issues (mostly related to local id vs global id)
- it looks like prespawning has a bug when we have multiple clients that share the same client id? It's because the client is using the connection-local ClientId instead of the global ClientId that is used everywhere in the server (and should be used inside NetworkConfig)
  - That seems annoying... should we just refuse the connection if there is a client id conflict?
  - or we need to find a way for the client to access the global ClientId. Maybe we should be sending a message when the client gets connected that includes the GlobalId. 
    - how does a client discover the client-ids of other clients? with a similar message?
    - should we just create a Global resoure on the client that contains some data such as the global client id?
    - maybe an easy way to achieve this is to create on the server one Entity per Client connected, with a component that contains the main data that we want to replicate. And then figure out rules for who this entity should be replicated to.
    

Closes https://github.com/cBournhonesque/lightyear/issues/115


.cargo/config.toml
```diff
@@ -1,15 +0,0 @@
-[build]
-#target = "wasm32-unknown-unknown"
-#target = "aarch64-apple-darwin"
-rustflags = ["--cfg", "web_sys_unstable_apis"]
```
Nul-led, line 4: what about the cfg flag?

examples/bullet_prespawn/src/main.rs
```diff
@@ -30,109 +32,101 @@ use lightyear::connection::netcode::{ClientId, Key};
 use lightyear::prelude::TransportConfig;
 use lightyear::shared::log::add_log_layer;
 
-// Use a port of 0 to automatically select a port
-pub const CLIENT_PORT: u16 = 0;
-pub const SERVER_PORT: u16 = 5000;
 pub const PROTOCOL_ID: u64 = 0;
```
Nul-led, line 35: can't we move this into the cfg file too?

examples/bullet_prespawn/src/main.rs
```diff
@@ -30,109 +32,101 @@ use lightyear::connection::netcode::{ClientId, Key};
...
 }
 
-cfg_if::cfg_if! {
-    if #[cfg(target_family = "wasm")] {
-        fn main() {
-            // NOTE: clap argument parsing does not work on WASM
+fn main() {
+    cfg_if::cfg_if! {
+        if #[cfg(target_family = "wasm")] {
+            std::panic::set_hook(Box::new(console_error_panic_hook::hook));
```
Nul-led, line 108: doesnt bevy automatically add the console error panic hook?

examples/certificates/cert.pem
```diff
@@ -1,11 +1,11 @@
 -----BEGIN CERTIFICATE-----
```
Nul-led, line 1: we should remove this file completely and only ever have the user generate their own

examples/certificates/key.pem
```diff
@@ -1,5 +1,5 @@
 -----BEGIN PRIVATE KEY-----
-MIGHAgEAMBMGByqGSM49AgEGCCqGSM49AwEHBG0wawIBAQQga38//86g0SxU++MI
-USzwolTXEHJcI8mHE9q1PL1HnEihRANCAAQbzFFi0ouIIjojztfw7TkVmjBx/DtS
-tUDM/TvggHWbrthlHhy91hpGwZtn+87da6QhjTzgjMIlccOzItriocOB
+MIGHAgEAMBMGByqGSM49AgEGCCqGSM49AwEHBG0wawIBAQQg519TDzD9PTgZndCB
```
Nul-led, line 2: we should remove this file completely and only ever have the user generate their own

lightyear/Cargo.toml
```diff
@@ -132,6 +132,7 @@ tokio-tungstenite = { version = "0.21.0", optional = true, features = [
 console_error_panic_hook = { version = "0.1.7" }
 ring = { version = "0.17.7", optional = true }
 web-sys = { version = "0.3", optional = true, features = [
+  "FileReader",
```
Nul-led, line 135: this can be removed in favor of include_str! i think

.cargo/config.toml
cBournhonesque: Ah right.. maybe i'll add it to the examples directly

examples/bullet_prespawn/src/main.rs
cBournhonesque: Yes but i was getting errors before bevy was loading (while parsing the settings). I can probably remove it

examples/certificates/cert.pem
cBournhonesque: I'd prefer to keep it for now to avoid users the friction of having to generate their own

PR #182, opened 2024-03-15
steam works

zwazel:
Just a small pull request, I added the `SingleClient`. As it needs to be threadsafe i wrapped it in a `Arc<RwLock<_>>` and implemented `Sync` and `Send`.
I'm not that experienced with Threadsafe stuff, so let me know if this is the correct way to do it.


cBournhonesque:
That looks good to me, thank you so much for the help!
I might end up just using directly `Arc<RwLock<_>>` for clarity instead of a wrapper.

PR #184, opened 2024-03-15
Add Reflect everywhere, use bevy::reflect::TypePath instead of Named

cBournhonesque:
## Add Reflect everywhere

I want to add Reflect for several reasons:
- it contains some convenience utilities such as getting the variant name of an enum, or in general getting the stable path of a name
- it looks like the ecosystem will use Reflection for the editor so it might be a good idea to start using it
- it has potential to allow me to split the protocol into multiple pieces, and use reflection, instead of having a giant enum for the MessageProtocol and ComponentProtocol
- allows for debugging via bevy_inspector_egui

Still, we have a big breaking change, which is that every protocol-related type (`Channel`,`Message`,`Input`) now need to implement `Reflect`.

TODO: add reflect for each resource/component and register them

## Remove the `Named` trait

Now that all types are `Reflect`, we can get rid of the `Named` trait and just use `reflect::TypePath`


Nul-led:
This is amazing :)
Lmk if u want a review

cBournhonesque:
Closing for now as the `Named` trait is not needed anymore (the message name is taken from the protocol directly)

PR #188, opened 2024-03-17
Add benchmark for packet packing using the new bitcode version

cBournhonesque:
Thanks to @finnbear for contributing an awesome benchmark on using bitcode 0.6.0 and packing the resulting serialized messages into packets of size MAX_SIZE.

The reason for this benchmark is that with the previous bitcode version 0.5.0; we could serialize each individual packet separately and track the number of bits written in an internal buffer. We could use this to try to serialize as many small messages as possible into packets of size MTU.

With bitcode 0.6.0, some APIs related to counting the number of bits written in the internal buffer have been removed; but it's still possible to apply several message-packing strategies to serialize various messages into packets of size MTU.

image

The benchmark shows that the best strategy is to use the `BinarySearchPacker`;
it only has a 2X time overhead compared to the naive packer (which doesn't do any packing) but it seems to very rarely create extra fragments beyond the MTU size.

### DIFF
--- a/benches/Cargo.toml
+++ b/benches/Cargo.toml
@@ -18,8 +18,14 @@ crossbeam-channel = "0.5.10"
 anyhow = { version = "1.0.75", features = [] }
 bevy = { version = "0.13", features = ["bevy_core_pipeline"] }
 derive_more = { version = "0.99", features = ["add", "mul"] }
-divan = "0.1.11"
+divan = "0.1.14"
 serde = { version = "1.0.188", features = ["derive"] }
+
+bitcode = "0.6.0-beta.1"
+rand = "0.8.5"
+rand_chacha = "0.3.1"
+lz4_flex = { version = "0.11.2", default-features = false }
+
 
 [[bench]]
 name = "spawn"
--- a/benches/bitcode_packing.rs
+++ b/benches/bitcode_packing.rs
@@ -1,0 +1,258 @@
+//! Benchmark for how to use bitcode to pack messages into packets of MTU bytes
+#![allow(unused_variables)]
+use bevy::app::App;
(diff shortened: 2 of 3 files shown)


caibear:
> but it seems to very rarely create extra fragments beyond the MTU size.

It only does this when individual messages are beyond the MTU size. That benchmark contains 3 of those messages (at least with the current rng seed). That's why even appended packer also has 3 extension packets.

> BinarySearchPacker

A more accurate name for this would be interpolation search. I was initially going to use a binary search which @finnbear communicated to you but I settled on [this more efficient algorithm](https://en.wikipedia.org/wiki/Interpolation_search).

cBournhonesque:
You're right, I renamed it to InterpolationPacker!

> It only does this when individual messages are beyond the MTU size. That benchmark contains 3 of those messages (at least with the current rng seed). That's why even appended packer also has 3 extension packets.

There's no guarantee for this right? Especially if `MAX_ATTEMPS` is low and `SAMPLE` is too high it could be possible that we get a packet that is bigger than MTU size.

caibear:
> There's no guarantee for this right?

Probably not, though I've never observed this behavior since `largest_so_far` usually has a value <=  `MAX_SIZE`.

caibear:
I've come up with a new packer that has better worst case performance and won't create any extraneous extension packets.
https://gist.github.com/caibear/68a9faf3f9a4ce94f321c7e06d2ea0ca#file-main-rs-L200-L281

It's 30% slower than interpolation packer if the message size is small and uniform, but it's 10x faster and generates half the packets on degenerate cases such as 10% messages larger MTU.

cBournhonesque:
Awesome, thanks!

PR #191, opened 2024-03-17
use select instead of channel to notify about closed task

df51d:


lightyear/src/transport/websocket/server.rs
```diff
@@ -87,7 +87,7 @@ impl Transport for WebSocketServerSocket {
                     let clientbound_tx_map = clientbound_tx_map.clone();
                     let serverbound_tx = serverbound_tx.clone();
                     IoTaskPool::get()
-                        .spawn(Compat::new(async move {
```
df51d, line 90: Already in the scope of async runtime, so it is unnecessary. This whole thing is closure passed to `Compat::new` above.

cBournhonesque:
Thanks, this is great!

PR #204, opened 2024-03-26
Fix broken links in book

msvbg:
Add `mdbook-linkcheck` and fix broken links in book.


cBournhonesque:
Awesome, thank you very much!

PR #205, opened 2024-03-27
Run client and server inside the same app

cBournhonesque:
# Objective

In the current "listen-server" mode, we create 2 Apps (client_app, server_app) that communicate via channels.

This is already pretty good but there are a couple of issues:
- there is a slight input delay even if the latency should be 0 because the inputs from the client will get received by the server on a later tick. Also the client timeline might still be a little different than the server timeline
- there could be an overhead from running 2 Apps, 2 Worlds with similar archetypes, etc.

Instead we could have a model where the `ClientPlugin` and `ServerPlugin` are both running inside the same `App`.

Ideally, there would be as little code change as possible between this mode and the other ways of running lightyear. (dedicated server)
Which means that in a first phase, we will keep the Interpolated/Predicted/Confirmed components even though they aren't actually needed:
- there is no need for prediction, the client could just move directly the server-authoritative entity. (but maybe the client could predict other players? see RocketLeague)
- there is no need for interpolation. Interpolation exists because the server updates are sporadic. In this case the client and server are in the same app so the client has access to the server state every frame!

# Design

I guess there's 2 main options:
A) This is what we'll do in this PR. We create both plugins in the same app connected via local channels. The main changes compared to 2 apps are:
- there is only one TimeManager that is shared between the client and the server . This means that the client operates on the SAME timeline as the server, instead of being ahead (for client-prediction)
- We cannot update the TimeManager's time in the SyncManager because the timeManager's time is the client's prediction time but also the server's time! Also we don't need to do it since the goal of SyncManager is to keep the two timelines in sync, and in this case there are perfectly in sync. We also need to adjust the interpolation time code.
- Change the ClientGlobalMetadata code to only use the client_id received from the server (i.e. without Replicate component)
- For inputs to work with 0 latency, we will send inputs directly via channels to the server's InputEvents receiver (otherwise there would be a one-frame delay because the inputs would be received on the next frame. Also having inputs not being received instantly on the server is incompatible with the unified timeline, since if that's the case the client would have to run slightly ahead of the server.
So basically I have to either:
- separate the TimeManager into a ClientTimeManager and a ServerTimeManager
- or force inputs to have 0 latencies in this unified mode, and use the same timeline.
I think the second option is easier to implement and has other benefits (true 0-delay input latency if running in singleplayer) so let's try this first. (we can keep the second option as fallback).

This has another consequence: **even if there is a link conditioner adding delay, the local client will always have 0 input latency from client -> server.**

B)
- do not use io at all between the server/client, and instead have the client **reuse directly the server entities**
- if a client has ShouldBePredicted, instead of spawning a new Predicted entity we just add a Predicted component
- if a client has ShouldBeInterpolated, instead of spawning a new Interpolated entity, we just add an Interpolated component to the server entity
- either the client doesn't connect to the App (no io needed since they are sharing the same world). This might need a couple of adjustments in the example code.
The problem is that there are some places in the code that expect a client-id, and maybe also a connection?
Can investigate this; maybe we need to create a fixed ClientId for the 'Unified' usecase.
This might be a good long term solution, but might be hard to get working in the short-term (need to let the app run with no io, no netcode, etc.)
I guess in this mode almost nothing related to the ClientPlugin is needed? no need for io, prediction, interpolation, etc. Maybe we could just run with the ServerPlugin? In which case the changes needed are:
- server creates the entity directly (not when a client connects! Maybe we can emit a connection event for a "local" client)
- inputs need to affect that entity

# Blockers

- [ ] Update all examples
- [x] Refactor TimeManager so that it can be shared between ClientPlugin and ServerPlugin
- [x] Register SharedPlugin only once if both ClientPlugin and SharedPlugin are called
- [x] For inputs, do NOT network the client inputs to the server if we are in the same app, instead just directly write to `ServerEventInputs` so that the inputs are read on the same tick by the server
  - [ ] check if it works for leafwing
- [X] Update the `SyncManager` to work correctly
- [ ] Disable sending pings if running in unified mode?

BUGS:
- [ ]: when another client connects, the interpolated entity in the unified app has no components
- [ ]: the interpolation delay on other clients seems very long!

### DIFF
--- a/examples/simple_box/src/client.rs
+++ b/examples/simple_box/src/client.rs
@@ -72,9 +72,7 @@ impl Plugin for ExampleClientPlugin {
 }
 
 // Startup system for the client
-pub(crate) fn init(mut commands: Commands, mut client: ResMut<ClientConnection>) {
-    commands.spawn(Camera2dBundle::default());
-
+pub(crate) fn init(mut client: ResMut<ClientConnection>) {
     let _ = client.connect();
 }
 
@@ -98,10 +96,16 @@ pub(crate) fn handle_connection(mut commands: Commands, metadata: Res<GlobalMeta
 // System that reads from peripherals and adds inputs to the buffer
 pub(crate) fn buffer_input(
     tick_manager: Res<TickManager>,
+    metadata: Res<GlobalMetadata>,
     mut connection_manager: ResMut<ClientConnectionManager>,
+    server_input_event: Option<ResMut<Events<server::InputEvent<Inputs>>>>,
     keypress: Res<ButtonInput<KeyCode>>,
 ) {
+    let Some(client_id) = metadata.client_id else {
+        return;
+    };
     let tick = tick_manager.tick();
+    let mut input = Inputs::None;
     let mut direction = Direction {
         up: false,
         down: false,
@@ -121,16 +125,24 @@ pub(crate) fn buffer_input(
         direction.right = true;
     }
     if !direction.is_none() {
-        return connection_manager.add_input(Inputs::Direction(direction), tick);
+        input = Inputs::Direction(direction);
     }
     if keypress.pressed(KeyCode::Backspace) {
-        // currently, inputs is an enum and we can only add one input per tick
-        return connection_manager.add_input(Inputs::Delete, tick);
+        input = Inputs::Delete;
     }
     if keypress.pressed(KeyCode::Space) {
-        return connection_manager.add_input(Inputs::Spawn, tick);
+        input = Inputs::Spawn;
     }
-    return connection_manager.add_input(Inputs::None, tick);
+    // if we run in listen-server mode and the server/client are running in the same app
+    // emit the input event directly to the server
+    if let Some(mut server_input_event) = server_input_event {
+        info!(?tick, "send client input");
+        // TODO: should we still add the event to the buffer, for rollbacks?
+        //  maybe not, because there cannot be mispredictions?
+        let _ = server_input_event.send(server::InputEvent::new(Some(input), client_id));
+    } else {
+        return connection_manager.add_input(input, tick);
+    }
 }
 
 // The client input only gets applied to predicted entities that we own
--- a/examples/simple_box/src/main.rs
+++ b/examples/simple_box/src/main.rs
@@ -14,15 +14,18 @@ use bevy::prelude::*;
 use bevy::DefaultPlugins;
 use bevy_inspector_egui::quick::WorldInspectorPlugin;
 use clap::{Parser, ValueEnum};
+use lightyear::prelude::client::{InterpolationConfig, InterpolationDelay};
 use serde::{Deserialize, Serialize};
 
 use lightyear::prelude::TransportConfig;
 use lightyear::shared::log::add_log_layer;
 use lightyear::transport::LOCAL_SOCKET;
 
-use crate::client::ClientPluginGroup;
-use crate::server::ServerPluginGroup;
+use crate::client::{ClientPluginGroup, ExampleClientPlugin};
+use crate::protocol::protocol;
+use crate::server::{ExampleServerPlugin, ServerPluginGroup};
 use crate::settings::*;
+use crate::shared::{shared_config, SharedPlugin};
 
 mod client;
 mod protocol;
@@ -67,6 +70,7 @@ fn main() {
 
 fn run(settings: Settings, cli: Cli) {
     match cli {
+        // ListenServer using a single app
         #[cfg(not(target_family = "wasm"))]
         Cli::ListenServer { client_id } => {
             // create client app
@@ -77,7 +81,7 @@ fn run(settings: Settings, cli: Cli) {
                 recv: from_server_recv,
                 send: to_server_send,
             };
-            let net_config = build_client_netcode_config(
+            let client_net_config = build_client_netcode_config(
                 client_id.unwrap_or(settings.client.client_id),
                 // when communicating via channels, we need to use the address `LOCAL_SOCKET` for the server
                 LOCAL_SOCKET,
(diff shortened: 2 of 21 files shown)


examples/simple_box/src/client.rs
```diff
@@ -98,10 +96,16 @@ pub(crate) fn handle_connection(mut commands: Commands, metadata: Res<GlobalMeta
 // System that reads from peripherals and adds inputs to the buffer
 pub(crate) fn buffer_input(
     tick_manager: Res<TickManager>,
+    metadata: Res<GlobalMetadata>,
     mut connection_manager: ResMut<ClientConnectionManager>,
+    server_input_event: Option<ResMut<Events<server::InputEvent<Inputs>>>>,
```
msvbg, line 101: Ideally it's transparent to clients whether they are publishing events in-app or if they do it over the network

examples/simple_box/src/client.rs
```diff
@@ -121,16 +125,24 @@ pub(crate) fn buffer_input(
...
-        return connection_manager.add_input(Inputs::Spawn, tick);
+        input = Inputs::Spawn;
+    }
+    // if we run in listen-server mode and the server/client are running in the same app
+    // emit the input event directly to the server
+    if let Some(mut server_input_event) = server_input_event {
+        info!(?tick, "send client input");
+        // TODO: should we still add the event to the buffer, for rollbacks?
+        //  maybe not, because there cannot be mispredictions?
+        let _ = server_input_event.send(server::InputEvent::new(Some(input), client_id));
```
msvbg, line 142: As above, I think it's probably the nicest if there's one abstraction that captures both cases

examples/simple_box/src/main.rs
```diff
@@ -77,27 +81,52 @@ fn run(settings: Settings, cli: Cli) {
...
+        //     let extra_transport_configs = vec![TransportConfig::Channels {
+        //         // even if we communicate via channels, we need to provide a socket address for the client
+        //         channels: vec![(LOCAL_SOCKET, to_server_recv, from_server_send)],
+        //     }];
+        //     let mut server_app = server_app(settings, extra_transport_configs);
+        //
+        //     // run both the client and server apps
+        //     std::thread::spawn(move || server_app.run());
+        //     client_app.run();
+        // }
```
msvbg, line 129: No need for this anymore!

lightyear/src/client/connection.rs
```diff
@@ -423,7 +423,7 @@ impl<P: Protocol> ReplicationSend<P> for ConnectionManager<P> {
         target: NetworkTarget,
         system_current_tick: BevyTick,
     ) -> Result<()> {
-        // trace!(?entity, "Send entity spawn for tick {:?}", self.tick());
+        info!(?entity, "Send entity spawn to server");
```
msvbg, line 426: This should probably still be a trace log

lightyear/src/client/input.rs
```diff
@@ -248,7 +248,10 @@ fn prepare_input_message<P: Protocol>(
     if !message.is_empty() {
         // TODO: should we provide variants of each user-facing function, so that it pushes the error
         //  to the ConnectionEvents?
-        debug!("sending input message: {:?}", message.end_tick);
+        info!(
+            ?current_tick,
+            "sending input message: {:?}", message.end_tick
+        );
```
msvbg, line 254: Should probably remain `debug!`

lightyear/src/client/plugin.rs
```diff
@@ -73,22 +73,26 @@ impl<P: Protocol> Plugin for ClientPlugin<P> {
...
             .add_plugins(ClientDiagnosticsPlugin::<P>::default());
+
+        if config.client_config.replication.enable {
+            app.add_plugins(ClientReplicationPlugin::<P>::new(tick_duration));
+        }
+
+        // check if are running both client and server plugins in the same app
+        if !app.is_plugin_added::<SharedPlugin<P>>() {
+            app.add_plugins(SharedPlugin::<P> {
+                config: config.client_config.shared.clone(),
```
msvbg, line 93: Is `client_config.shared` always the same as `server_config.shared`?

lightyear/src/client/replication.rs
```diff
@@ -3,9 +3,16 @@ use bevy::utils::Duration;
...
 use crate::client::connection::ConnectionManager;
 use crate::client::sync::client_is_synced;
+use crate::prelude::client::InterpolationDelay;
 use crate::prelude::{Protocol, ReplicationSet};
 use crate::shared::replication::plugin::ReplicationPlugin;
 
+#[derive(Clone, Default, Debug)]
+pub struct ReplicationConfig {
+    /// Set to true to enable replicating this client's entities to the server
+    pub enable: bool,
```
msvbg, line 13: This has default `false` now, but `true` is maybe a better default

lightyear/src/server/plugin.rs
```diff
@@ -65,19 +65,23 @@ impl<P: Protocol> PluginType for ServerPlugin<P> {
...
+
+        if !config.server_config.replication.disable {
+            app.add_plugins(ServerReplicationPlugin::<P>::new(tick_duration));
+        }
+
+        // check if are running both client and server plugins in the same app
+        if !app.is_plugin_added::<SharedPlugin<P>>() {
+            app.add_plugins(SharedPlugin::<P> {
+                // TODO: move shared config out of server_config?
+                config: config.server_config.shared.clone(),
```
msvbg, line 82: Same question as earlier

lightyear/src/server/replication.rs
```diff
@@ -6,6 +6,13 @@ use crate::server::connection::ConnectionManager;
 use crate::server::prediction::compute_hash;
 use crate::shared::replication::plugin::ReplicationPlugin;
 
+/// Configuration related to replicating the server's World to clients
+#[derive(Clone, Default, Debug)]
+pub struct ReplicationConfig {
+    /// Set to true to disable replicating this server's entities to clients
+    pub disable: bool,
```
msvbg, line 13: Here it's called `disable`, and in the other one it's called `enabled`

examples/simple_box/src/client.rs
cBournhonesque: Yep, this is just for the proof-of-comcept; I'll try to hide it in the internal APIs

examples/simple_box/src/main.rs
cBournhonesque: I think I might still keep it as an example; maybe
`ListenServer` -> 2 apps 
`Unified` -> 1 app with both client and server

What do you think of the naming?

lightyear/src/server/replication.rs
cBournhonesque: That's on purpose; I think 99% of cases are only interested in server->client replication. Client->server replication is a bit more niche; and will need further work. I probable need to create a new component or field `replication_direction`

lightyear/src/client/plugin.rs
cBournhonesque: Yes

lightyear/src/client/connection.rs
cBournhonesque: Yes, the PR is still WIP :) 
Sorry I should have made it clearer

examples/simple_box/src/main.rs
msvbg: Hmm yeah naming is hard. Unified/segregated might be a good split? I don't really know

PR #207, opened 2024-03-28
Run client and server in same app with unified timeline

cBournhonesque:
# Objective

In the current "listen-server" mode, we create 2 Apps (client_app, server_app) that communicate via channels.

This is already pretty good but there are a couple of issues:
- there is a slight input delay even if the latency should be 0 because the inputs from the client will get received by the server on a later tick. Also the client timeline might still be a little different than the server timeline
- there could be an overhead from running 2 Apps, 2 Worlds with similar archetypes, etc.

Instead we could have a model where the `ClientPlugin` and `ServerPlugin` are both running inside the same `App`.

Ideally, there would be as little code change as possible between this mode and the other ways of running lightyear. (dedicated server)
Which means that in a first phase, we will keep the Interpolated/Predicted/Confirmed components even though they aren't actually needed:
- there is no need for prediction, the client could just move directly the server-authoritative entity. (but maybe the client could predict other players? see RocketLeague)
- there is no need for interpolation. Interpolation exists because the server updates are sporadic. In this case the client and server are in the same app so the client has access to the server state every frame!

# Design

I guess there's 2 main options:
A) This is what we'll do in this PR. We create both plugins in the same app connected via local channels. The main changes compared to 2 apps are:
- there is only one TimeManager that is shared between the client and the server . This means that the client operates on the SAME timeline as the server, instead of being ahead (for client-prediction)
- We cannot update the TimeManager's time in the SyncManager because the timeManager's time is the client's prediction time but also the server's time! Also we don't need to do it since the goal of SyncManager is to keep the two timelines in sync, and in this case there are perfectly in sync. We also need to adjust the interpolation time code.
- Change the ClientGlobalMetadata code to only use the client_id received from the server (i.e. without Replicate component)
- For inputs to work with 0 latency, we will send inputs directly via channels to the server's InputEvents receiver (otherwise there would be a one-frame delay because the inputs would be received on the next frame. Also having inputs not being received instantly on the server is incompatible with the unified timeline, since if that's the case the client would have to run slightly ahead of the server.
So basically I have to either:
- separate the TimeManager into a ClientTimeManager and a ServerTimeManager
- or force inputs to have 0 latencies in this unified mode, and use the same timeline.
I think the second option is easier to implement and has other benefits (true 0-delay input latency if running in singleplayer) so let's try this first. (we can keep the second option as fallback).

This has another consequence: **even if there is a link conditioner adding delay, the local client will always have 0 input latency from client -> server.**

B)
- do not use io at all between the server/client, and instead have the client **reuse directly the server entities**
- if a client has ShouldBePredicted, instead of spawning a new Predicted entity we just add a Predicted component
- if a client has ShouldBeInterpolated, instead of spawning a new Interpolated entity, we just add an Interpolated component to the server entity
- either the client doesn't connect to the App (no io needed since they are sharing the same world). This might need a couple of adjustments in the example code.
The problem is that there are some places in the code that expect a client-id, and maybe also a connection?
Can investigate this; maybe we need to create a fixed ClientId for the 'Unified' usecase.
This might be a good long term solution, but might be hard to get working in the short-term (need to let the app run with no io, no netcode, etc.)
I guess in this mode almost nothing related to the ClientPlugin is needed? no need for io, prediction, interpolation, etc. Maybe we could just run with the ServerPlugin? In which case the changes needed are:
- server creates the entity directly (not when a client connects! Maybe we can emit a connection event for a "local" client)
- inputs need to affect that entity

# Changes
- Make `TimeManager` be a shared resource created in SharedPlugin
- Add `unified` field in `SharedConfig`
- Update Sync to work with a unified timeline
- make sure `TimeManager::update()` is called only once, even in unified mode
- Split out the input_buffer in a separate InputManager, and handle unified mode separately (send inputs via a channel)
- add a `ReplicationConfig` to specify in which direction the replication will be (a better long-term solution will be to add a ReplicationDirection component)
- Update client GlobalMetadata to work correctly in unified mode.

# Blockers

- [ ] Update all examples
  - [x] Simple box
  - [x] Replication groups
  - [x] priority
  - [ ] leafwing inputs: BUG
  - [ ] interest management: interpolation timeline is wonky on the unified server
  - [x] client replication
  - [ ] bullet prespawn: problem with prespawn, and also with controls
- [x] Refactor TimeManager so that it can be shared between ClientPlugin and ServerPlugin
- [x] Register SharedPlugin only once if both ClientPlugin and SharedPlugin are called
- [x] For inputs, do NOT network the client inputs to the server if we are in the same app, instead just directly write to `ServerEventInputs` so that the inputs are read on the same tick by the server
  - [X] check if it works for leafwing
- [X] Update the `SyncManager` to work correctly
- TODO: maybe in unified mode, we add a component ReplicatedFromServer/ReplicatedFromClient to identify entities?

BUGS:
- [ ]: when another client connects, the interpolated entity in the unified app has no components
- [ ]: the interpolation delay on other clients seems very long!


cBournhonesque:
I'll just merge this for now as there are too many changes in this PR to be manageable, but I think that running the ClientPlugin and the ServerPlugin in the same world for now is not manageable:
- a Replicate entity is replicated both from client->server and server->client so we have to add extra components ReplicateToClientOnly and ReplicateToServerOnly
- the replication system-sets have to be separated between client and server
-all queries have to be adjusted to keep in mind that client entities and server entities are living in the same world
- the examples have shown that it's very hard to be able to have client and server in the same world without significant mental loaded added

This discord thread: https://discord.com/channels/691052431525675048/747940465936040017/1223411454334734436 by @msvbg outlines 3 potential approaches for client and server on the same machine:
- A) 2 separate apps (unity way, also what the current 'listen-server' mode does with local channels as io
- B) 1 world with both Client and Server Plugin (what this PR does, but which proves to be too much of a footgun)
- C) 1 world with only the Server Plugin, but the server also acts as a player

I'll merge this now and focus my effort and getting C) working. B) would work only when bevy supports multiple worlds in the same app

Issue #211, opened 2024-03-31
Listen-server with only Server plugin


cBournhonesque:
Continuation of https://github.com/cBournhonesque/lightyear/pull/207:

Instead of having the client and server entities in the same world, we will try to get a listen-server working only with the Server plugin.

Potential changes:
- Maybe no lightyear client plugin? or just a plugin with some most functionality disabled
- No networking to the local client, but maybe use a ClientId::LocalClient to refer to the local client?
- Inputs
  - leafwing: just add the InputMap directly on the server entity when it spawns.
  - native: enable a ClientPlugin that just writes directly to the server receive input events without buffering
- Replication
  - no replication to the local client since there's no networking, the local client can directly see the server world
  - other players' entities that are replicated and have InterpolationTarget excluding the local client will have Interpolated added.
  - when spawning the local player's entity, add Predicted as component
- Messages
  - how do other players send a message to the host server? just with send_message to server? no NetworkTarget?
  
  
STATUS:
- I still include the client plugin but i disable the replication/prediction/interpolation/networking plugins
- I still create a ConnectionManager on the client with a ClientId of type ClientId::Local, and I create a fake connection of type Local which does nothing, but enables the user to connect/disconnect
- For native inputs:
  - we forward the input_manager directly to the server inputs
  - we don't forward to the client inputs (to avoid inputs getting handled twice)
- For leafwing:
  - we also forward to server, but maybe we don't need to, and directly act with the client?
- I add Predicted/Interpolated on the server-side depending on the value of the replication target

EXAMPLES:
- simple_box: works
- replication_group: works
- priority: works
- interest_management: works, but interest management doesn't work for the local client (which is expected)
- leafwing: works
- client_replication: works
- bullet-pre-spawn: works

LEARNINGS
- Use `Or<With<Predicted>, With<Replicate>>` rather than `<Without<Confirmed>, Without<Interpolated>>` to target Predicted + Server entities. The second option doesn't work in HostServer mode because other client's entities will have 'interpolated' added to them!

TODO:
- pre-prediction might not work because:
  - the server expects to receive a replicated component from the client (to then replicate it back to the original client) but in a host-server situation the entity is just directly created.
- leafwing: maybe same?
- we get warnings like: 
```
2024-04-01T02:32:23.109836Z  WARN lightyear::shared::replication::send: Trying to insert a component that is already in the message group_id=ReplicationGroupId(4294967303) entity=7v1 kind=ShouldBePredicted
```

PR #212, opened 2024-04-01
Cb/listen server server only

cBournhonesque:
Fixes https://github.com/cBournhonesque/lightyear/issues/211


examples/client_replication/src/client.rs
```diff
@@ -52,23 +52,26 @@ pub(crate) fn init(mut client: ResMut<ClientConnection>) {
...
-                ..default()
-            },
-        ));
-        // spawn a local cursor which will be replicated to other clients, but remain client-authoritative.
-        commands.spawn(CursorBundle::new(
-            client_id,
-            Vec2::ZERO,
-            color_from_id(client_id),
-        ));
+        if config.shared.mode == Mode::Separate || client_id.is_local() {
```
msvbg, line 59: Which cases are excluded from this condition? My brain is having trouble parsing this.

examples/client_replication/src/client.rs
```diff
@@ -132,14 +135,19 @@ fn player_movement(
...
+/// Spawn a server-owned pre-predicted player entity when the space command is pressed
 fn spawn_player(
     mut commands: Commands,
     mut input_reader: EventReader<InputEvent<Inputs>>,
     connection: Res<ClientConnection>,
     players: Query<&PlayerId, With<PlayerPosition>>,
+    config: Res<ClientConfig>,
 ) {
     let client_id = connection.id();
+    if config.shared.mode == Mode::HostServer && !client_id.is_local() {
```
msvbg, line 147: Can this be replaced by some sort of `has_authority` run condition like in replicon?

lightyear/src/client/input.rs
```diff
@@ -140,43 +141,52 @@ impl<P: Protocol> Plugin for InputPlugin<P> {
...
-                )
-                    .chain(),
-            );
+        // SYSTEMS
+        app.add_systems(
+            FixedPostUpdate,
+            clear_input_events::<P::Input>.in_set(InputSystemSet::ClearInputEvent),
+        );
+
+        if app.world.resource::<ClientConfig>().shared.mode == Mode::HostServer {
```
msvbg, line 150: I think it might be better for this to be controlled via run conditions so it can be toggled at runtime

lightyear/src/client/input.rs
```diff
@@ -330,3 +331,19 @@ fn prepare_input_message<P: Protocol>(
     input_manager.input_buffer.pop(interpolation_tick);
     // .pop(current_tick - (message_len + 1));
 }
+
+/// In host server mode, we don't buffer inputs (because there is no rollback) and we don't send
+/// inputs through the network, we just send directly to the server's InputEvents
+fn send_input_to_server<A: UserAction>(
```
msvbg, line 337: Maybe call this `*_unbuffered` or something so it doesn't sound like the standard code path

lightyear/src/server/replication.rs
```diff
@@ -80,6 +83,14 @@ impl<P: Protocol> Plugin for ServerReplicationPlugin<P> {
                 (compute_hash::<P>
                     .in_set(InternalReplicationSet::<ServerMarker>::SetPreSpawnedHash),),
             );
+
+        if app.world.resource::<ServerConfig>().shared.mode == Mode::HostServer {
```
msvbg, line 87: Same concern as above about run conditions

lightyear/src/shared/config.rs
```diff
@@ -17,14 +17,32 @@ pub struct SharedConfig {
...
+
+#[derive(Default, Clone, Copy, Debug, PartialEq, Eq)]
+pub enum Mode {
+    #[default]
+    /// Run the client and server in two different apps
+    Separate,
+    /// Run the client and server in the same bevy app/world
+    /// NOTE: this is complicated to get right, and is not recommended, as it is hard to
+    /// distinguish between client entities and server entities
+    Unified,
```
msvbg, line 31: Should this be removed given that it's unworkable?

examples/client_replication/src/client.rs
cBournhonesque: Yes I think this is actually not needed: the issue was that there is some code handling client-side ConnectEvents and some code handling server-side ConnectEvents.

In this PR I had merged the 2 into a single one, so even when another client than the local client was connecting, it was also triggering the client-side ConnectEvent code. I will separate the two events and this change probably won't be needed anymore

examples/client_replication/src/client.rs
cBournhonesque: As mentioned above, this is fixed by separating the ConnectEvents.

I do think that it will be necessary to have more control in lightyear over who has authority. I see that unreal has more finegrained control over 'ControlledPawns', etc.
It's definitely something I want to look into

lightyear/src/client/input.rs
cBournhonesque: You're right, thanks

lightyear/src/shared/config.rs
cBournhonesque: Yes, I think i'll remove it

lightyear/src/client/input.rs
cBournhonesque: Yep, definitely, having things more run-time configurable is the next thing I will work on. I'll just keep things like this for now since everything is very static, and I'll work on the runtime configurability in a separate PR

PR #223, opened 2024-04-03
Cb/improve transport

cBournhonesque:


lightyear/src/transport/wrapper/mod.rs
```diff
@@ -0,0 +1,15 @@
+//! Module defining 'wrappers' that modify the behaviour of an existing [`PacketReceiver`] or [`PacketSender`].
+//!
+//! Wrappers are used to add additional functionality to an existing transport, such as encryption, compression, metrics, etc.
```
msvbg, line 3: A lot of http servers call this "middleware", which could be an alternative name.

lightyear/src/transport/mod.rs
```diff
@@ -72,8 +77,50 @@ pub trait PacketReceiver: Send + Sync {
...
+}
+
+impl PacketReceiver for &mut dyn PacketReceiver {
+    fn recv(&mut self) -> Result<Option<(&mut [u8], SocketAddr)>> {
+        (*self).recv()
+    }
+}
+
+impl PacketReceiver for Box<&mut dyn PacketReceiver> {
+    fn recv(&mut self) -> Result<Option<(&mut [u8], SocketAddr)>> {
```
msvbg, line 123: `Result<Option<...>>` looks a little weird to me

lightyear/src/connection/netcode/server.rs
```diff
@@ -474,7 +474,7 @@ impl<Ctx> NetcodeServer<Ctx> {
         &mut self,
         addr: SocketAddr,
         packet: Packet,
-        sender: &mut impl PacketSender,
+        sender: &mut dyn PacketSender,
```
msvbg, line 477: Why `dyn` instead of `impl`? Is it to keep the types simpler?

lightyear/src/transport/io.rs
```diff
@@ -251,36 +221,44 @@ pub struct IoStats {
...
-    pub fn stats(&self) -> &IoStats {
-        &self.stats
+    fn split(&mut self) -> (&mut (dyn PacketSender + '_), &mut (dyn PacketReceiver + '_)) {
+        // todo: compression + bandwidth monitoring
+        let (sender, receiver) = self.transport.split();
+        if let Some(conditioner) = &mut self.conditioner {
+            self.receiver = Some(Box::new(ConditionedPacketReceiver::new(
+                receiver,
+                conditioner,
+            )));
```
msvbg, line 257: Feels a little wrong that this mutates `self`

msvbg (commented):
I hope you don't find my reviews annoying!

lightyear/src/transport/mod.rs
cBournhonesque: It's because I want to differentiate between:
- error: io error
- Ok(None): no error, but no packets
- Ok(Some()): no error, and there is a packet.

(this part did not change btw)

lightyear/src/connection/netcode/server.rs
cBournhonesque: Not too sure, I think the compiler was shouting at me :)

lightyear/src/transport/wrapper/mod.rs
cBournhonesque: yes i like that!

cBournhonesque:
Closing in favor of https://github.com/cBournhonesque/lightyear/pull/222

PR #238, opened 2024-04-12
Updated title.md in tutorial to not derive `Message` trait anymore.

schnippl0r:
In the actual example code, the `Message` trait is not derived, so I removed it from the source code listing in the tutorial.


cBournhonesque (approved):
Thank you!

PR #311, opened 2024-05-05
Support bevy_xpbd with double precision

msvbg:
I use bevy_xpbd with f64 scalars, which doesn't work with lightyear currently since it assumes f32. I initially tried making `LerpFn` generic over the float type, but this gets a little hairy and I figured that it was easier to just keep lightyear's base interpolation logic as f32 and convert convert f32 to f64 as needed in the lerp utilities.


cBournhonesque (approved):
LGTM, thanks!

PR #321, opened 2024-05-07
Clarify client replication guide

philpax:
Spent a bit of time debugging why my client-replicated entity didn't show up on the server and discovered it was because of the client/server configs, so I figured I'd make a note of them in the guide to make it more obvious. (Might be a good idea to share the harness between the examples so that it's easier to see what's different between them?)

I've also added an example of how to spawn an entity with replication on the client, just to illustrate that there's no magic, but I'm happy to remove it if it's redundant. I did notice
> Note that `prediction_target` and `interpolation_target` will be unused as the server doesn't do any 
> prediction or interpolation. 

directly after where I inserted the code; however, I pulled those values for the fields from here:
https://github.com/cBournhonesque/lightyear/blob/23a1ed82371c370ce82db8c9ca9a8b99115fdae4/examples/client_replication/src/protocol.rs#L54-L58

Not sure if they're actually necessary or not?

### DIFF
--- a/book/src/concepts/advanced_replication/client_replication.md
+++ b/book/src/concepts/advanced_replication/client_replication.md
@@ -1,12 +1,54 @@
 # Client replication
 
 It is also possible to use lightyear to replicate entities from the client to the server.
-There are different possibilities.
+There are different possibilities. Note that regardless of the approach taken, you will
+need to explicitly enable sending replication from the client, and receiving replication
+on the server:
+
+```rust
+let client_config = client::ClientConfig {
+    /* other fields */
+    replication: lightyear::client::replication::ReplicationConfig {
+        enable_send: true,
+        enable_receive: true,
+    },
+    ..default()
+};
+
+let server_config = server::ServerConfig {
+    /* other fields */
+    replication: lightyear::server::replication::ReplicationConfig {
+        enable_send: true,
+        enable_receive: true,
+    },
+    ..default()
+};
+```
+
+For a detailed example, please see the `client_replication` example.
 
 ## Client authoritative
 
 To replicate a client-entity to the server, it is exactly the same as for a server-entity.
-Just add the `Replicate` component to the entity and it will be replicated to the server.
+Just add the `Replicate` component to the entity and it will be replicated to the server:
+```rust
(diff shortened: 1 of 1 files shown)


cBournhonesque:
Thanks, yes the `enable_send`/`enable_receive` are a huge footgun, I actually got stuck because of that more than once.
I might get rid of them actually; I think they were meant as an optimization to avoid having useless replication-send systems on the client, but they cna be a big footgun.

Thanks for the PR, i'll take a look soon.
And yes maybe it would be nice to have a common harness for the tests, it would certainly make updates easier (compared with manually updating 8 tests...). Although I do like that the tests are currently all completely independent and work as self-contained examples.

philpax:
> Thanks, yes the `enable_send`/`enable_receive` are a huge footgun, I actually got stuck because of that more than once.
> I might get rid of them actually; I think they were meant as an optimization to avoid having useless replication-send systems on the client, but they cna be a big footgun.

Makes sense! If the overhead of running the systems is minimal, I'd leave them on. You might still want to have them around so that the user can explicitly specify whether they want to accept replication from the client - it is, after all, a security concern to let the client replicate up arbitrary entities. You'd just need to make it obvious.

> And yes maybe it would be nice to have a common harness for the tests, it would certainly make updates easier (compared with manually updating 8 tests...). Although I do like that the tests are currently all completely independent and work as self-contained examples.

I definitely appreciate the standalone aspect, but I found the existing per-test harness to be a bit too dense; I had to scroll back and forth through `simple_box`'s `main.rs` and cross-reference it with `client.rs` and `server.rs` to see exactly what I needed to integrate where. 

Setting up a common harness *might* help with that by only exposing what's relevant for that example and abstracting away the details of setting up support for each protocol and execution mode, but it could also add to the confusion by requiring another file that you have to consult. It's a tough balance to walk 😅

msvbg:
This is a little off-topic but I do find the examples a little bit hard to follow as it stands because there's so much going on and they support so many different modes of operation (lightyear is rich in features!). I think ideally examples would be smaller and contained in one file only so that they could be run using `cargo run --example`.

PR #323, opened 2024-05-08
Use sparse set storage for corrections

msvbg:
It looks like `Correction` is added and removed frequently, which means that it should probably use [sparse set ECS storage](https://bevy-cheatbook.github.io/patterns/component-storage.html) to stop the ECS from moving entities between archetypes too frequently.

I'm not really sure how to bench this just yet due to not being experienced enough with the code base, but this probably deserves a benchmark.


cBournhonesque:
Yes, we need better benchmarks; in the meantime I agree that correction should get added/removed at every rollback, which could be frequent. I'll merge this for now.

Issue #243, opened 2024-04-15
Issues with steam when we recreate the ClientConnection multiple times


cBournhonesque:
### Issues with `NetworkingConfigEntry`
There are several issues with `NetworkingConfigEntry`: 
- https://github.com/Noxime/steamworks-rs/pull/168
- https://github.com/Noxime/steamworks-rs/issues/169

### Issues when recreating the ClientConnection
  - I tried dropping the existing ClientConnection (which calls steam api's shutdown) before re-creating the new one. Looks like this doesn't work because steams "shutdown" api pretty much does nothing; steam api is running until the program is shutdown: https://github.com/rlabrecque/Steamworks.NET/issues/187#issuecomment-323700244)
  - I tried initializing the steamworks clients only once via OnceCell, the client is able to connect, but is then disconnected immediately with `reason: RemoteBadCert`
    - I tried adding a background system that just runs steam callbacks at all times even if we're not connected, but that didn't help either
    - in the file 'cef_log.previous`, i See logs saying `Error parsing cert retrieved from AIA (as DER). Couldn't read tbsCertificate as SEQUENCE. Failed parsing Certificate`
    - If I connect immediately when starting the app, it works (for some unknown reason)
    - If I don't connect immediately; on the first connection attempt I get disconnected with an error `RemoteBadCert`, but future connection attempts work!

PR #346, opened 2024-05-12
Support steam p2p sockets

msvbg:
Adds support for Steam P2P sockets. This has been tested by @zwazel and is confirmed to at least work partially, though with some kinks (https://github.com/cBournhonesque/lightyear/issues/243).

I didn't add tests for this since that would require Steam to be installed to run tests.

### DIFF
--- a/lightyear/src/connection/steam/mod.rs
+++ b/lightyear/src/connection/steam/mod.rs
@@ -5,8 +5,6 @@ pub(crate) mod client;
 pub(crate) mod server;
 pub(crate) mod steamworks_client;
 
-pub use steamworks_client::SteamworksClient;
-
 pub(crate) fn get_networking_options(
     conditioner: &Option<LinkConditionerConfig>,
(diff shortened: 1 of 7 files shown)


lightyear/src/connection/steam/mod.rs
```diff
@@ -3,13 +3,7 @@ use steamworks::networking_types::{NetworkingConfigEntry, NetworkingConfigValue}
 
 pub(crate) mod client;
 pub(crate) mod server;
-
-// NOTE: it looks like there's SingleClient can actually be called on multiple threads
-// - https://partner.steamgames.com/doc/api/steam_api#SteamAPI_RunCallbacks
-pub(crate) struct SingleClientThreadSafe(steamworks::SingleClient);
```
cBournhonesque, line 9: So do you think it would be possible to use this as a wrapper to the user's provided `SteamworksClient` to avoid the `Arc<RwLock>`?

cBournhonesque:
Looks great to me, thanks!

PR #354, opened 2024-05-15
Better rollback debug logs

msvbg:
Slightly improved debug logging here. I am now keeping debug logs on for rollback at all times to spot issues. I guess this will become untenable since this is a perfectly normal code path to hit.


cBournhonesque:
Thank you!

PR #355, opened 2024-05-15
Remove preprediction override from bullet prespawn example

msvbg:


cBournhonesque:
Actually upon double-checking the player entities are pre-predicted. You can see that the `PrePredicted` component is added on the client when the entity gets spawned on the client.

Issue #144, opened 2024-02-12
WASM examples broken if the user switches tabs


cBournhonesque:
I think this PR broke the examples for some reason.
- the examples work fine on native
- on wasm, the connection gets timed out pretty quickly for some reason

UPDATE:
- it looks like the update is that the client tasks stop running when the page is alt-tabbed, so the connection times out if the user alttabs for too long!
- the second problem is that the server seems to be stuck in a weird state where it doesn't accept webtransport connections anymore.

On client we get:
```
Failed to establish a connection to https://127.0.0.1:5000/: net::ERR_QUIC_PROTOCOL_ERROR.QUIC_NETWORK_IDLE_TIMEOUT (No recent network activity after 9003077us. Timeout:9s).
failed to connect to server: Error(JsValue(WebTransportError: Opening handshake failed.Error: Opening handshake failed.))
```

POSSIBLE SOLUTIONS:
- run in background thread using webworkers?

Nul-led:
@cBournhonesque apparently the thread does not actually get stopped entirely but instead just throttled. Might be possible to figure out if that happens and temporarily stop sending and receiving packets.

Disabling ram saver seems to work on brave, cant say with other browsers. Requires more testing ig.

simbleau:
It's very unclear what the real issue is from reading this.

Can someone confirm:
- Who causes the disconnection? Does the server terminate the client, does the client terminate themselves, does the browser terminate the session, etc.?

Regarding the "Keep Alive" strategy:
- I don't think this is possible. Unlike websockets which are browser-scoped, WebTransport is tab-scoped. That means when you switch tabs, WebTransport fully stops. The KeepAlive strategy only works for websockets.
- However, could we instead modify the constraint to allow messages an infinite amount of time to return a response, to prevent timeout?

Other strategies:
- Would automatic re-connection be possible?
- Has anyone tried a web worker to keep it alive? This seems like it would work, since the script will run independently of the main thread, allowing it to continue executing even when the user switches tabs.

MOZGIII:
If you have audio playing on your tab it won't get suspended. It would actually be quite fine for a game to apply this workaround.

Automatic reconnection at the WebTransport layer and all that is not possible since the browser doesn't really give us control over those details of the connection that would enable it: specifically, timeouts. It is the browser that would terminate the WebTransport session, and this will happen regardless of whether tab is paused or not, so we can't hook into it.

[...]

PR #371, opened 2024-05-22
Add WebPlugin to keep running in background

cBournhonesque:
Fixes https://github.com/cBournhonesque/lightyear/issues/144

All credit for this PR goes to @Nul-led who came up with the idea and did the entire implementation. 
I just copy-pasted his code to put it in a `WebPlugin`. Thank you very much for this amazing solution!

In wasm, the main thread gets quickly throttled by the browser when it is hidden (e.g. when the user switches tabs).
This means that the app.update() function will not be called, because bevy's scheduler only runs `app.update()` when
the browser's requestAnimationFrame is called. (and that happens only when the tab is visible)

This is problematic because:
- we stop sending packets so the server disconnects the client because it doesn't receive keep-alives
- when the client reconnects, it also disconnects because it hasn't been receiving server packets
- the internal transport buffers can overflow because they are not being emptied

This solution spawns a WebWorker (a background thread) which is not throttled, and which runs
`app.update()` at a fixed interval. This way, the client can keep sending and receiving packets,
and updating the local World.

TODO:
- probably don't need `world.clear_trackers()`
- make interval configurable?
- confirm that the WebWorker doesn't get throttled even after long periods?


simbleau:
This has universally great functionality - Should be a plugin one can easily `cargo add webworker_update` and add low-frequency updates to their application

lightyear/src/client/web.rs
```diff
@@ -0,0 +1,54 @@
...
+    )
+    .unwrap();
+
+    let worker = Worker::new(&Url::create_object_url_with_blob(&blob).unwrap()).unwrap();
+
+    let closure = Closure::<dyn FnMut()>::new(move || {
+        if window().unwrap().document().unwrap().hidden() {
+            // Imitate app.update()
+            let world = unsafe { world_ptr.write().unwrap().as_mut().unwrap() };
+            world.run_schedule(Main);
```
simbleau, line 46: Magnifique

Nul-led:
@simbleau thanks :p
ill probably make a crate for it then

Nul-led:
re: configurable interval
1) we use format! to make it configurable at startup
2) we make an actual protocol for changing the duration (could be achieved by replacing setInterval with setTimeout)

Nul-led:
re: trottled web workers
doesnt seem to happen (tested over 30 minutes), didnt confirm on webkit since i dont have a mac (feel free to check if anyone does), confirmed that it works for at least 7 minutes on android chromium based (more than that doesnt seem reasonable anyways)

Nul-led:
Another note ill save here:
Lightyear doesnt need to send inputs while in background mode, so maybe worth looking into to save bandwidth?

Nul-led:
> > @simbleau thanks :p ill probably make a crate for it then
> 
> Please let me know when it's done, I'd add this to my bevy_rtc project today.

maybe make an impl urself for now then, ill prob wont have time until weekend

PR #375, opened 2024-05-23
Draw Confirmed entity outlines in xpbd_physics demo

RJ:
i find it useful to see where the server confirmed entities are in relation to the predicted ones:

https://github.com/cBournhonesque/lightyear/assets/29747/4240eba6-74b4-4258-ab40-3212f7d5c229

feel free to close if you don't want it in the demo though


cBournhonesque:
Thanks :) ! I'll add it for now, but i'll probably put this behind a setting in the settings file

PR #457, opened 2024-06-05
Arena allocator

cBournhonesque:
# Context
There are a lot of situations where we suddenly need to allocate a bunch of small objects that we will then discard:
- when buffering all the updates that will go into the replication messages
- when we do priority filtering
- when build the final packets
- when we read the received packets into data structures

It might be beneficial to use an fast arena allocator that allocated a bunch of memory. We use this allocator in the hot path, and then regularly we just cheaply drop the chunk of memory at once.

# Solution
Introduces a `ArenaManager` that wraps a `SyncBlinkAlloc` from the blink-alloc crate, which provides a multi-thread safe arena allocator.

Use this arena everytime we send to allocate the `pending_actions` and `pending_updates` that we need to build the final `ReplicationMessage`.

# Benchmarks
`send_insert_float/1_client`

| num_entities | main | branch |
| --- | --- | --- |
| 1000 | 1.30ms | 1.21ms |
| 10000 | 13.10ms | 13.23ms |

Again it doesn't seem to be bringing much improvement, maybe the allocations in replication_send are not the problem, rather it's the ones in the packet building steps?
Double-check.

Anyways it's nice to have a working implementation


msvbg (commented):
All of the unsafe transmutes makes me a little uneasy! I would suggest encapsulating the unsafety somewhere and adding a `// SAFETY` comment to explain why it's safe

cBournhonesque:
> All of the unsafe transmutes makes me a little uneasy! I would suggest encapsulating the unsafety somewhere and adding a `// SAFETY` comment to explain why it's safe

This doesn't actually work right now, I'm just trying to get a proof of concept working so that I can really benchmark.
I think it would be easier to get this working if all the replication-send systems were combined into one, or if the `ReplicationSend` became a `NonSend` resource

PR #480, opened 2024-06-18
New example: Spaceships

RJ:
New example called spaceships. 

* Client spaceships are spawned upon connect, and despawned when a client disconnects.
* Early inputs result in predicted bullet spawning to reduce perceived lag
* Visual smoothing and error correction used to blend in any rollback mispredictions
* All entities are predicted (ie, in your local timeline, ahead of server) so collisions between dynamic bodies should be non-janky
* Player labels: `25±2ms [3]` means 25ms server reported ping, 2ms jitter, 3 ticks of future inputs available

Made a few small quality of life and visibility changes in lightyear.

Native and wasm:

spaceships-screenshot


lightyear/src/client/prediction/prespawn.rs
```diff
@@ -177,6 +178,10 @@ impl PreSpawnedPlayerObjectPlugin {
                                 trace!(?kind, "using kind for hash");
                                 kind.hash(&mut hasher)
                             });
+                            // if a user salt is provided, hash after the sorted component list
```
cBournhonesque, line 181: The salt is to avoid hash collisions between multiple entities?

lightyear/src/client/prediction/prespawn.rs
```diff
@@ -374,15 +380,52 @@ impl PreSpawnedPlayerObjectPlugin {
...
     /// The hash that will identify the spawned entity
     /// By default, if the hash is not set, it will be generated from the entity's archetype (list of components) and spawn tick
     /// Otherwise you can manually set it to a value that will be the same on both the client and server
     pub hash: Option<u64>,
+    /// An optional extra value that will be passed to the hasher as part of the default hashing algorithm
+    ///
+    /// Since the default hash uses the tick and components, a useful addition is the client id, to
+    /// distinguish between bullets spawned on the same tick, but by different players.
+    #[serde(skip)]
+    pub user_salt: Option<u64>,
```
cBournhonesque, line 407: Do you think I should always add the client_id automatically?

lightyear/src/server/connection.rs
```diff
@@ -462,6 +462,16 @@ impl Connection {
         }
     }
 
+    /// Return the latest estimate of rtt
+    pub fn rtt(&self) -> Duration {
```
cBournhonesque, line 466: Maybe I should just make ping_manager public? but yes these are convenient

examples/spaceships/assets/fonts/quicksand-light.txt
```diff
@@ -0,0 +1,93 @@
+Copyright 2011 The Quicksand Project Authors (https://github.com/andrew-paglinawan/QuicksandFamily), with Reserved Font Name “Quicksand”.
```
cBournhonesque, line 1: Is this needed? we can probably do without this in the example

cBournhonesque (approved):
I only looked at the non example code, looks good overall! I'll look at the rest later. 
Thank you so much, the example is awesome

lightyear/src/client/prediction/prespawn.rs
RJ: if two players fire on the same tick, the hashes would otherwise collide because the bullets would have identical archetypes and tick.

lightyear/src/client/prediction/prespawn.rs
RJ: probably a good idea, although when non player characters both spawn bullets on the same tick, you'd still want to add a salt with the NPC id too, for example

lightyear/src/server/connection.rs
RJ: slight preference towards a few nice fns in Connection, since there is internal stuff in pingmanager that games aren't concerned with. rtt(), jitter(), last_ping_time() or so. probably not many 🤷

examples/spaceships/assets/fonts/quicksand-light.txt
RJ: let me figure out if i can use a bevy built-in font to avoid bundling this.. i found the player labels to be really useful when debugging so don't want to forgo labels entirely.

lightyear/src/client/prediction/prespawn.rs
RJ: actually the client_id shouldn't be default added because that would make it impossible to predict spawns of global stuff, since the server wouldn't be using your client id.

RJ: consider the case of a cloud spawning raindrops every 3 seconds. clients could predict those, but if clients include their client id in the hash, it won't match the server. the cloud / raindrops don't belong to a player, so there is no client id to include on the server side.

examples/spaceships/assets/fonts/quicksand-light.txt
RJ: changed to using the built-in font and removed this

PR #501, opened 2024-06-25
Spaceships example: bullet collisions and predict missing inputs

RJ:
* Bullet collisions added: collision events used for visuals on client, manipulating score on server.
* Predict missing remote player inputs

### DIFF
--- a/examples/spaceships/src/client.rs
+++ b/examples/spaceships/src/client.rs
@@ -196,6 +196,7 @@ fn player_movement(
         if let Some((prev_tick, prev_input)) = input_buffer.get_last_with_tick() {
             let staleness = (tick - prev_tick).max(0) as u16;
             if staleness > MAX_STALE_TICKS {
+                // input too stale, apply default input (ie, nothing pressed)
                 apply_action_state_to_player_movement(
                     &ActionState::default(),
                     staleness,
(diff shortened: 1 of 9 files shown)


examples/spaceships/src/client.rs
```diff
@@ -146,13 +148,65 @@ fn handle_new_player(
...
+                    &ActionState::default(),
+                    staleness,
+                    &mut aiq,
+                    tick,
+                );
+            } else {
+                apply_action_state_to_player_movement(prev_input, staleness, &mut aiq, tick);
+            }
+        } else {
+            apply_action_state_to_player_movement(action_state, 0, &mut aiq, tick);
```
cBournhonesque, line 209: why do we have this here? we just apply the latest received ActionState?

cBournhonesque (approved):
LGTM!
The bloom and explosions look really good
I might have to move this from a `examples` folder to a `demo` folder as it is becoming quite big :)

RJ:
> LGTM! The bloom and explosions look really good I might have to move this from a `examples` folder to a `demo` folder as it is becoming quite big :)

i convinced myself that the additional visuals are either instructive, or will help to debug network issues..

examples/spaceships/src/client.rs
cBournhonesque: I meant l209. Thanks, the comments are helpful

PR #506, opened 2024-06-29
Remove WebTransport use directives from WebSocket module

blinkdog:
I was working on something using the WebSocket transport, and noticed a compile error.

The WebSocket transport module seems to import, but does not use, some things from WebTransport.

I think it may be a copypasta from the WebTransport module that just wasn't caught because both features were enabled (in my project, only the websocket feature is enabled).


cBournhonesque:
Indeed, it looks like it! Thank you

Nul-led:
Ah my bad probably 😂

PR #539, opened 2024-07-16
Rename EventMessage to more specific name

notmd:
I found myself often have to rename these to avoid import conflict, also when accepting autocomplete I have to look for which module it comes from.
- Rename `MessageEvent` in the client module to `ClientMessageEvent` and in the server module to `ServerMessageEvent` and deprecate the old names to make migration easier.

### DIFF
--- a/examples/avian_physics/src/server.rs
+++ b/examples/avian_physics/src/server.rs
@@ -103,7 +103,7 @@ pub(crate) fn movement(
 
 pub(crate) fn replicate_inputs(
     mut connection: ResMut<ConnectionManager>,
-    mut input_events: EventReader<MessageEvent<InputMessage<PlayerActions>>>,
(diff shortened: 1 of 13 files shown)


cBournhonesque:
It's actually intentional that they have the same name, because normally I would expect your code to be separated into a `client` module and a `server` module.
You would import `lightyear::prelude::client::*` in one and `lightyear::prelude::server::*` in the other, so there would be no conflict.

Can you give an example of what you're trying to do which requires importing both at the same time?

notmd:
I don't split client and server in my code. I split them by feature instead. So both client code and server are in one file.

cBournhonesque:
Maybe I can an additional section in the prelude where I given an alias to all the structs that have a conflicting name, that seems like a good compromise

PR #571, opened 2024-08-07
Update book visual interp page with fixed physics example

RJ:
Updated the visual interp book page. key insight is this bit:

Avian's `SyncPlugin` does this for you, but **beware if you use FixedUpdate**, you need to run the `SyncPlugin` in `PostUpdate`, otherwise you won't be syncing changes correctly. Visual interp happens in `PostUpdate`, even if physics runs in `FixedUpdate`.

```rust
// Run physics in FixedUpdate, but run the SyncPlugin in PostUpdate
app
  .add_plugins(
    PhysicsPlugins::new(FixedUpdate)
      .build()
      .disable::<SyncPlugin>(),
  )
  .add_plugins(SyncPlugin::new(PostUpdate));
```

### DIFF
--- a/book/src/concepts/advanced_replication/visual_interpolation.md
+++ b/book/src/concepts/advanced_replication/visual_interpolation.md
@@ -16,27 +16,17 @@ To solve this, `lightyear` provides the `VisualInterpolation` plugin.
 The plugin will take care of interpolating the position of the entity between the last two `FixedUpdate` ticks, thus making sure that 
 the entity is making smooth progress on every frame.
 
-There are 3 main ways we can interpolate the component:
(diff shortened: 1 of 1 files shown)


book/src/concepts/advanced_replication/visual_interpolation.md
```diff
@@ -16,27 +16,17 @@ To solve this, `lightyear` provides the `VisualInterpolation` plugin.
 The plugin will take care of interpolating the position of the entity between the last two `FixedUpdate` ticks, thus making sure that 
 the entity is making smooth progress on every frame.
 
-There are 3 main ways we can interpolate the component:
```
cBournhonesque, line 19: I'd prefer to keep these 3 methods in the write-up; but yes maybe we can make it clearer that lightyear uses the first one

cBournhonesque (approved):

PR #578, opened 2024-08-10
Fixed Local client not sending Connect event in HostServer Mode + Added Test

zwazel:


lightyear/src/tests/host_server_stepper.rs
```diff
@@ -247,11 +247,10 @@ impl HostServerStepper {
         self.client_app
             .world_mut()
             .run_system_once(|mut commands: Commands| commands.connect_client());
-
         // Advance the world to let the connection process complete
         for _ in 0..100 {
             if self
-                .client_app
+                .server_app
```
cBournhonesque, line 253: this should be `self.client_app`, because the external client needs to check if they're synced

cBournhonesque (commented):
I have one comment but other this LGTM; thanks!

lightyear/src/tests/host_server_stepper.rs
zwazel: yep, originally didnt want to push this change, just overlooked it. fixed now :)

cBournhonesque:
Thanks, great addition :)

PR #596, opened 2024-08-16
Excessive iteration over query in simple_box example

nazgul72:
While debugging inputs in HostServer mode it seems excessive iteration of position_query is done when the input is actually none. 

I know this is example code, but programmers (at least one;me) always copy paste, so probably good to fix anyway.


cBournhonesque:
Good catch, thanks!

PR #598, opened 2024-08-17
Fix Pre-Prediction

cBournhonesque:
- PrePrediction was broken since the introduction of send-side entity mapping.
That's because:
- client spawns predicted E1 and replicates it to server.
- server spawns E2 and maintains mapping E2<>E1 and replicates it to client
- client spawns confirmed E3 and maintains mapping E2<>E3.

With sender-side mapping, updates related to E2 would be mapped to E1 before being sent.

This PR fixes this problem, and it also works in host-server mode and with hierarchy.

Also fixes:
- the `replication_groups` example was failing because of a erroneous check in the protocol
- the `connect`/`disconnect` systems were running twice (both the normal and the host-server versions were running)
- the server doesn't replicate `Spawns` to the entity that first spawned an entity.

### DIFF
--- a/NOTES.md
+++ b/NOTES.md
@@ -1,52 +1,19 @@
-# Entity mapping
+# PrePrediction
 
-- SOLUTION 0: the entity receiver always keeps the mapping.
-  - we always try to apply the mapping from remote->local on receive side
-  - we always try to apply the mapping from local->remote on send side.
-  - C1 spawns E1, sends to S, S spawns E0. S replicated to C2 who spawns E2
-    C1 sends a message about E1, entity gets mapped to E0.
-    S rebroadcasts to C1, mapping from E0 to E1.
-    S rebroadcasts to C2, no mapping. The receiver maps E0 to E2.
-  - We know there are no conflicts because only the receiver has the mapping!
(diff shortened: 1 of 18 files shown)


cBournhonesque:
My solution was to do:
- spawn E1-predicted on client with PrePredicted
- on client, an observer detects the addition of PrePredicted and spawns Confirmed E2
- on server, as soon as the PrePredicted entity is received (into a server entity E3), we transfer authority back to server, and we update the mapping so that the server has E3<>E2 mapping.

The issue is that the `client.ReplicationReceiver.local_entities` is not updated correctly since the server doesn't spawn E2 normally...

Maybe a solution would be to keep what was done before with a slight adjustement:
- client spawns E1 with PrePredicted and replicates to server. As soon as the message is sent, PrePredicted is removed.
- server spawns E2 and has a mapping from E2<>E1.
  - **NEW**: we manually remove the E2<>E1 mapping so that server messages about E2 aren't converted to E1
- server replicates E2 to client. Client creates E3 with PrePredicted (and maintains E3<>E2 mapping). Using PrePredicted, client does the prediction mapping between E3 and E1

PR #599, opened 2024-08-17
Spaceships example: fix Visual Interpolation bits

RJ:
Clarifies how the visual interp is set up, and fixes the syncplugin issue as described in the lightyear book VisualInterpolation chapter

This removes the stutter from the text labels that follow the players, since they are positioned with `Transform`, not `Position`

Also filters collisions between players and their own bullets, because it was possible to collide with your own bullets while accelerating.

### DIFF
--- a/examples/spaceships/src/renderer.rs
+++ b/examples/spaceships/src/renderer.rs
@@ -45,8 +45,8 @@ impl Plugin for SpaceshipsRendererPlugin {
         app.add_systems(
             Last,
             (
-                add_player_visual_components,
-                update_player_visual_components,
+                add_player_label,
+                update_player_label,
                 draw_walls,
                 draw_confirmed_shadows.run_if(move || draw_shadows),
                 draw_predicted_entities,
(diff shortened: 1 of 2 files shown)


RJ:
`cargo test` passes for me locally, not sure why it's failing here

examples/spaceships/src/renderer.rs
```diff
@@ -63,10 +63,48 @@ impl Plugin for SpaceshipsRendererPlugin {
...
+// (Note also that we've configured avian's SyncPlugin to run in PostUpdate)
+fn add_visual_interpolation_components<T: Component>(
+    trigger: Trigger<OnAdd, T>,
+    q: Query<Entity, (With<T>, Without<Wall>, Without<Confirmed>)>,
+    mut commands: Commands,
+) {
+    if !q.contains(trigger.entity()) {
+        return;
     }
+    info!("Adding visual interp component to {:?}", trigger.entity());
```
cBournhonesque, line 101: We might want this log as debug

cBournhonesque (approved):
Thanks this looks great to me!

One of the unit tests is a bit flimsy

examples/spaceships/src/renderer.rs
RJ: agreed , changed

PR #606, opened 2024-08-19
Send events as server::events::InputEvent instead of client::events:InputEvent  in send_input_directly_to_client_events when using HostServer

nazgul72:
This change simplifies the HostServer mode by not requiring a separate system on the client side for HostServer. 

In apps that do not do client predictions they will be simpler, as there is no need for shared code.

There is also the added benefit that the system only copies input when doing local connection, and could in principal run on all clients (not sure if that's a great idea, but there it is)

I tested it in my app, and ran hostserver for  simple_box, not sure if there are any scenarios it would not work in.

### DIFF
--- a/lightyear/src/client/input/leafwing.rs
+++ b/lightyear/src/client/input/leafwing.rs
@@ -975,7 +975,7 @@ mod tests {
         assert!(stepper
             .client_app
             .world()
-            .entity(client_entity)
+            .entity(server_entity)
             .get::<ActionState<LeafwingInput1>>()
             .unwrap()
             .pressed(&LeafwingInput1::Jump));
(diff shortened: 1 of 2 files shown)


nazgul72:
I fixed the lint and format issue (I think), but it still fails on something I do not recognize. The test that fails seems very related, I need to look into that a bit more

lightyear/src/client/input/leafwing.rs
```diff
@@ -975,7 +975,7 @@ mod tests {
         assert!(stepper
             .client_app
             .world()
-            .entity(client_entity)
```
cBournhonesque, line 978: this should still be `client_entity`

cBournhonesque:
The failing test is unrelated; this works for me, thanks!

PR #607, opened 2024-08-20
Remove unneeded import

FastestMolasses:
If I don't include the web transport feature, then I get an error when building my project because this is imported on all configurations, and it's not used in the file.


Nul-led:
Lgtm

PR #608, opened 2024-08-23
Fix certificate generate commands in examples

qtfkwk:
* Prepended a `cd` command because all other commands in each README assume user is in that directory; also the `generate.sh` script assumes user is in the root directory; alternatively consider adding `cd "$(dirname "$(realpath "$0")")"` and removing `$OUT/` in the `generate.sh` and then it could be run from anywhere and the generated keypair would be there (?)
* Used `$(git rev-parse --show-toplevel)` instead of `..` or `../..` for consistency
* Corrected path from `examples/generate.sh` to `examples/certificates/generate.sh`


cBournhonesque:
Thanks, that's great!

PR #609, opened 2024-08-23
Fix listen-server commands in examples

qtfkwk:
AFAICT, the `listen-server` command was renamed to `client-and-server` in https://github.com/cBournhonesque/lightyear/commit/0f9947b24a81948bd6aac6cc1f6b1ac234b18dfb ... this PR intends to update all the dangling mentions of the no longer valid `listen-server` command.


cBournhonesque:
You're right, thank you!

PR #617, opened 2024-08-31
Create server-authoritative 3D physics example

nick-e:
I've created an example of clients controlling server-authoritative, 3D physics objects. The spaceships example also showcases clients controlling server-authoritative physics but it has a lot of extra features such as pre-spawning and wasm that make it more difficult to follow. This is intended to be a more streamlined example. Plus, it demonstrates 3D physics which is neat.

[avian_3d_physics_demo.webm](https://github.com/user-attachments/assets/4d89e18d-2f08-434d-9d3b-4c701afa5fa5)

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -21,9 +21,10 @@ This folder contains various examples that showcase various `lightyear` features
 
 ## Advanced
 
-- `xpbd_physics`: example that shows how to replicate a physics simulation using xpbd.
+- `avian_physics`: example that shows how to replicate a physics simulation using xpbd.
   We also use the `leafwing` feature for a better way to manage inputs.
-- `spaceships`: more advanced version of `xpbd_physics` with player movement based on forces, fully server authoritative, predicted bullet spawning. 
+- `avian_3d_character`: example that shows clients controlling server-authoritative 3D objects simulated using avian.
+- `spaceships`: more advanced version of `avian_physics` with player movement based on forces, fully server authoritative, predicted bullet spawning. 
 - `bullet_prespawn`: example that shows how to spawn player-objects on the Predicted timeline. This is useful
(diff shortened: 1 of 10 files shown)


cBournhonesque:
Thanks, that's really awesome!

- What settings did you use for the video? I tried `input_delay_ticks = 4` and the corrections were very noticeable, it was much better with `input_delay_ticks = 6`
- I notice that there are constant rollbacks happening. This can be fixed in a future PR, but do you have any idea why?

nick-e:
One window was `cargo run --release -- client-and-server` and the other was `cargo run --release -- client -c 1`. I did not change the settings from what's in the commit.

I wish I knew why there are constant rollbacks 😅 I tried incorporating https://github.com/Jondolf/avian/commit/63580d38a50322a1f80ba55925094db47749d68e but that didn't stop the rollbacks. I think the movement logic is responsible. When the character looks like it isn't moving it is still getting a non-zero, very small force applied to it in order to keep it in place. That may result in a very small linear velocity that isn't in sync with the server. I was hoping that because the physics engine is deterministic that the resulting calculated movement force would also be deterministic but maybe not.

nick-e:
If I could get a snapshot of the world when the client resets the world to a tick when performing a rollback and also get a snapshot of the world on the server at that tick then I could then compare the two snapshots and see what's different. Is that possible?

nick-e:
Even better would be a snapshot of the world on the starting rollback tick and the tick after that plus the same ticks on the server. Then I could verify how the first tick changed.

cBournhonesque:
Usually I just print out the value of Position at every tick on the client and server and look for discrepancies.
In general avian is deterministic so it might be something else

nick-e:
That is much easier, thanks!

Issue #618, opened 2024-09-01
Incorrect time resource value during rollbacks


nick-e:
`bevy::prelude::Time::delta_seconds()` is incorrect when called within a system using the `FixedUpdate` schedule during a rollback. I expect it to be equal to the fixed time set when initializing the `Time` resource:
```rust
// `delta_seconds()` should then return 1.0 / 64.0 which is equal to 0.015625.
app.insert_resource(Time::new_with(Physics::fixed_once_hz(
   64.0
)));
```

# Replication steps:
The `avian_3d_physics` example runs at a tick rate of 64 Hz which means that `bevy::prelude::Time::delta_seconds()` should always return `0.015625` within a system that is in the `FixedUpdate` schedule. However, I noticed that when a rollback starts `delta_seconds()` will return a very small number (.e.g. `0.006358229`). The number is different every time.

1. Go to `examples/avian_3d_character/src/client.rs` and add the following to the start of `handle_character_actions`:
```rust
fn handle_character_actions(
    time: Res<Time>,
    spatial_query: SpatialQuery,
    mut query: Query<
        (
            &ActionState<CharacterAction>,
            &InputBuffer<CharacterAction>,
            CharacterQuery,
        ),
        With<Predicted>,
    >,
    tick_manager: Res<TickManager>,
    rollback: Option<Res<Rollback>>,
) {
    if let Some(ref rb) = rollback {
        if let Some(tick) = rb.get_rollback_tick() {
            info!("    rb-{:?}, {}", tick, time.delta_seconds());
        } else {
            info!("not-rb-{:?}, {}", tick_manager.tick(), time.delta_seconds());
        }
    } else {
        info!("{:?}, {}", tick_manager.tick(), time.delta_seconds());
    }
    
   ...
}
```
`handle_character_actions` is a system that is run using the `FixedUpdate` schedule.

2. Go to `examples/avian_3d_character` and run  `cargo run --release -- client-and-server`.
3. Compare the rollback times to the non-rollback times:
```
2024-09-01T03:16:09.523359Z  INFO avian_3d_character::client: not-rb-Tick(103), 0.015625
2024-09-01T03:16:09.527650Z  INFO avian_3d_character::client:     rb-Tick(103), 0.00417795
2024-09-01T03:16:09.540010Z  INFO avian_3d_character::client: not-rb-Tick(104), 0.015625
[...]

Issue #585, opened 2024-08-12
Handle non-networked rollback for resources


cBournhonesque:
We want to be able to rollback client resources whenever a rollback happens, so that the resource is reverted to a previous state.

Note that here we only consider resources that were not replicated from the server.

Similar to https://github.com/cBournhonesque/lightyear/issues/573 but for resources

PR #619, opened 2024-09-01
Set correct time resource for rollbacks

nick-e:
Create a new `Time<Fixed>` resource when `run_rollback()` performs a rollback. Set the elapsed time of the fixed time to the beginning of the rollback. Sadly, `avian_3d_character` still has constant rollbacks after this fix but the time delta and elapsed time are now correct for systems using the fixed update scheduler. I also noticed that the `run_rollback()` system may be executed even if `num_rollback_ticks` is zero.

Fixes https://github.com/cBournhonesque/lightyear/issues/618

### DIFF
--- a/examples/avian_3d_character/Cargo.toml
+++ b/examples/avian_3d_character/Cargo.toml
@@ -14,26 +14,26 @@ lightyear_examples_common = { path = "../common" }
 bevy_screen_diagnostics = "0.6"
 leafwing-input-manager = "0.15"
 avian3d = { version = "0.1.1", default-features = false, features = [
-    "3d",
-    "f32",
-    "parry-f32",
-    "parallel",
-    "serialize",
+  "3d",
+  "f32",
+  "parry-f32",
+  "parallel",
+  "serialize",
 ] }
 lightyear = { path = "../../lightyear", features = [
-    "webtransport",
-    "websocket",
-    "leafwing",
-    "avian3d",
+  "webtransport",
+  "websocket",
+  "leafwing",
+  "avian3d",
 ] }
 serde = { version = "1.0.188", features = ["derive"] }
 anyhow = { version = "1.0.75", features = [] }
 tracing = "0.1"
 tracing-subscriber = "0.3.17"
 bevy = { version = "0.14", features = [
-    "multi_threaded",
-    "bevy_state",
-    "serialize",
+  "multi_threaded",
+  "bevy_state",
+  "serialize",
 ] }
 
 rand = "0.8.1"
--- a/examples/avian_3d_character/assets/settings.ron
+++ b/examples/avian_3d_character/assets/settings.ron
@@ -1,6 +1,6 @@
 MySettings(
   server_replication_send_interval: 50,
-  input_delay_ticks: 4,
+  input_delay_ticks: 6,
   // do not set a limit on the amount of prediction
   max_prediction_ticks: 100,
   correction_ticks_factor: 2.0,
--- a/lightyear/src/client/prediction/predicted_history.rs
+++ b/lightyear/src/client/prediction/predicted_history.rs
@@ -404,7 +404,10 @@ mod tests {
         let confirmed = stepper
             .client_app
             .world_mut()
-            .spawn(Confirmed::default())
+            .spawn(Confirmed {
+                tick,
+                ..Default::default()
+            })
             .id();
         let predicted = stepper
             .client_app
@@ -548,10 +551,14 @@ mod tests {
         let mut stepper = BevyStepper::default();
 
         // add predicted, component
+        let tick = stepper.client_tick();
         let confirmed = stepper
             .client_app
             .world_mut()
-            .spawn(Confirmed::default())
+            .spawn(Confirmed {
+                tick,
+                ..Default::default()
+            })
             .id();
         let predicted = stepper
             .client_app
--- a/lightyear/src/client/prediction/rollback.rs
+++ b/lightyear/src/client/prediction/rollback.rs
@@ -9,6 +9,7 @@ use bevy::prelude::{
     Resource, With, Without, World,
 };
 use bevy::reflect::Reflect;
+use bevy::time::{Fixed, Time};
 use parking_lot::RwLock;
 use tracing::{debug, error, trace, trace_span};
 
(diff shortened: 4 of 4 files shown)


nick-e:
The only test that is failing is `test_buffer_inputs_with_delay` and that test will succeed if `--engine llvm` isn't specified. It appears that test has been failing before this PR.

cBournhonesque:
Yes you can ignore the input test failure!

So i'm trying to better understand what this will fix. I would have hoped that it would have fixed the avian3d constant rollbacks (but i guess not since there shouldn't be any rollbacks in the first place)


Basically the use-case is that it's possible that some other plugins or logic depends on `Time<Fixed>` (maybe physics) to perform some internal computations). Without this change, the `delta()` which indicates the delta between the last time the schedule ran is incorrect.

Note that with this change the `overstep_fraction()` might be incorrect after rollback? I'm wondering if it could impact visual interpolation or interpolation in general. Maybe it's safer to just rollback the entire resource?

nick-e:
In `avian_3d_character`, forces are applied to characters only in `apply_character_action()` and the physics engine itself. The force in `apply_character_action()` is calculated using the tick's delta seconds. Before this PR, delta seconds was incorrect for the first tick of the rollback which resulted in a force out of sync with the server. After this PR, the calculated force on the first tick of the rollback is in sync because delta seconds is in sync with the server. This has resulted in less visible jitter of the character on the client but the position/velocity of the character on the tick right after the first rollback is still very slightly out of sync with the server which triggers constant rollbacks.

I added some print statements to `apply_character_action()` to show that. Here are some server logs:
```
Tick(218), dt: 0.015625, pos: [-3.0433052, 1.2499869, 0.8792649], vel: [-3.3866599, -0.00000019290434, -1.7763437], force: [13.273917, 0, -12.635028]
Tick(219), dt: 0.015625, pos: [-3.094039, 1.249987, 0.85218054], vel: [-3.1512887, 0.0000069986404, -1.7930776], force: [12.844528, 0, -13.071289]
```

And here are some client logs:
```
    rb-Tick(218), dt: 0.015625, pos: [-3.0433052, 1.2499869, 0.8792649], vel: [-3.3866599, -0.00000019290434, -1.7763437], force: [13.273917, 0, -12.635028]
not-rb-Tick(219), dt: 0.015625, pos: [-3.094039, 1.249987, 0.8521804], vel: [-3.1512897, 0.000007011368, -1.7931191], force: [12.844612, 0, -13.071205]
```
`rb-tick` denotes that the tick is a rollback tick. `dt` is `Res<Time>::delta_seconds()`. `pos` and `vel` are the position and linear velocity of the character when `apply_character_action()` is called. `force` is the force to be applied to the character in response to the client's inputs. This force is based on the character's position and linear velocity. The logs show that the force applied to the character in the initial rollback tick is exactly the same as the server and yet the next tick shows that the character's linear velocity is just very slightly out of sync with the server. 

I'm not sure how to rollback the time resource completely because I'm not sure how to calculate the overstep of `Time<Fixed>` at the initial rollback tick. I think that this issue would need to be fixed first: https://github.com/cBournhonesque/lightyear/issues/585.

nick-e:
Closing the PR because it isn't a perfect fix. More info can be found here: https://github.com/cBournhonesque/lightyear/issues/618. I think lightyear needs to support rollback for non-network resources in order to correctly rollback the time resource instead of trying to do it manually in `run_rollback()`.

cBournhonesque:
This is how I added the rollback for the non-replicated components: https://github.com/cBournhonesque/lightyear/pull/583/files#diff-84f948fafa0de170ebd6f2acf3664f7d23f0da411653df88c5c7bc6baf85ab51R187

You need:
- a `ComponentHistory` (in this case a `ResourceHistory`) that stores the resource values at previous ticks.
- a `prepare_rollback` function that runs before the rollback actually happens, and that restores the resource value to the previous value (or despawns/spawns it if needed)

cBournhonesque:
> In `avian_3d_character`, forces are applied to characters only in `apply_character_action()` and the physics engine itself. The force in `apply_character_action()` is calculated using the tick's delta seconds. Before this PR, delta seconds was incorrect for the first tick of the rollback which resulted in a force out of sync with the server. After this PR, the calculated force on the first tick of the rollback is in sync because delta seconds is in sync with the server. This has resulted in less visible jitter of the character on the client but the position/velocity of the character on the tick right after the first rollback is still very slightly out of sync with the server which triggers constant rollbacks.
> 
> I added some print statements to `apply_character_action()` to show that. Here are some server logs:
> 
> ```
> Tick(173), dt: 0.015625, pos: [0.5802094, 1.2499866, 0.48336923], vel: [-3.0422049, 0.000006657445, 3.4980302], force: [-18.273233, 0, 1.3891554], dir: [-0.70710677, 0, 0.70710677]
> Tick(174), dt: 0.015625, pos: [0.53087044, 1.2499864, 0.5383434], vel: [-3.1999848, -0.0000053525696, 3.536802], force: [-18.32583, 0, -0.06926482], dir: [-0.70710677, 0, 0.70710677]
> ```
> 
> And here are some client logs:
> 
> ```
>     rb-Tick(173), dt: 0.015625, pos: [0.5802094, 1.2499866, 0.48336923], vel: [-3.0422049, 0.000006657445, 3.4980302], force: [-18.273233, 0, 1.3891554], dir: [-0.70710677, 0, 0.70710677]
> not-rb-Tick(174), dt: 0.015625, pos: [0.53110653, 1.2499864, 0.53836536], vel: [-3.199166, -0.000005247282, 3.5368834], force: [-18.325804, 0, -0.07351523], dir: [-0.70710677, 0, 0.70710677]
> ```
> 
> `rb-tick` denotes that the tick is a rollback tick. `dt` is `Res<Time>::delta_seconds()`. `pos` and `vel` are the position and linear velocity of the character when `apply_character_action()` is called. `dir` is the directional input used to control the character. `force` is the force to be applied to the character in response to the client's inputs. This force is based on `dt`, `dir`, `pos`, and `vel`. The logs show that the force applied to the character in the initial rollback tick is exactly the same as the server and yet the next tick shows that the character's position and linear velocity are just very slightly out of sync with the server.
> 
> I'm not sure how to rollback the time resource completely because I'm not sure how to calculate the overstep of `Time<Fixed>` at the initial rollback tick.

This is extremely helpful, thanks!
Why do you think there is discrepancy at the non-rb-tick?

nick-e:
I can try to make rollback-able resources. Those resources are a lot of help, thanks!

PR #622, opened 2024-09-04
Support rolling back resources

nick-e:
Support rollback for resources. Make the ReadyBuffer not require its item to implement PartialEq. ReadyBuffer should only have one item per tick and so only the tick needs to be compared and not the item. Call `App::add_resource_rollback<R>()` to add rollback support for a resource.

Fixes https://github.com/cBournhonesque/lightyear/issues/585

### DIFF
--- a/lightyear/src/client/prediction/plugin.rs
+++ b/lightyear/src/client/prediction/plugin.rs
@@ -1,6 +1,6 @@
 use bevy::prelude::{
     not, App, Component, Condition, FixedPostUpdate, IntoSystemConfigs, IntoSystemSetConfigs,
-    Plugin, PostUpdate, PreUpdate, Res, SystemSet,
+    Plugin, PostUpdate, PreUpdate, Res, Resource, SystemSet,
 };
 use bevy::reflect::Reflect;
 use bevy::transform::TransformSystem;
@@ -29,9 +29,10 @@ use crate::shared::sets::{ClientMarker, InternalMainSet};
 
 use super::pre_prediction::PrePredictionPlugin;
 use super::predicted_history::{add_component_history, apply_confirmed_update};
+use super::resource_history::update_resource_history;
 use super::rollback::{
     check_rollback, increment_rollback_tick, prepare_rollback, prepare_rollback_non_networked,
-    prepare_rollback_prespawn, run_rollback, Rollback, RollbackState,
+    prepare_rollback_prespawn, prepare_rollback_resource, run_rollback, Rollback, RollbackState,
 };
 use super::spawn::spawn_predicted_entity;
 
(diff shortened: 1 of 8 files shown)


lightyear/src/utils/ready_buffer.rs
```diff
@@ -15,28 +15,34 @@ pub struct ReadyBuffer<K, T> {
...
-impl<K: Ord + Clone, T: PartialEq> ReadyBuffer<K, T> {
+impl<K: Ord + Clone, T> ReadyBuffer<K, T> {
     /// Adds an item to the heap marked by time
     pub fn push(&mut self, key: K, item: T) {
         self.heap.push(ItemWithReadyKey { key, item });
     }
 
+    /// Returns a reference to the item with the highest `K` value or None if the queue is empty.
+    /// Does not remove the item from the queue.
+    pub fn peek_latest_item(&self) -> Option<(K, &T)> {
```
cBournhonesque, line 42: Wouldn't this return the item with the lowest K value?

lightyear/src/utils/ready_buffer.rs
```diff
@@ -160,23 +166,23 @@ pub struct ItemWithReadyKey<K, T> {
...
 }
 
-impl<K: Ord, T: PartialEq> Eq for ItemWithReadyKey<K, T> {}
+impl<K: Ord, T> Eq for ItemWithReadyKey<K, T> {}
 
-impl<K: Ord, T: PartialEq> PartialEq<Self> for ItemWithReadyKey<K, T> {
+impl<K: Ord, T> PartialEq<Self> for ItemWithReadyKey<K, T> {
     fn eq(&self, other: &Self) -> bool {
-        self.item == other.item && self.key == other.key
+        self.key == other.key
```
cBournhonesque, line 173: But what if we compare we have two component histories with values at the same ticks but the values are different?
Then we would get `ready_buffer_1 == ready_buffer_2`?

lightyear/src/client/prediction/plugin.rs
```diff
@@ -199,6 +200,17 @@ pub fn add_non_networked_rollback_systems<C: Component + PartialEq + Clone>(app:
     );
 }
 
+pub fn add_resource_rollback_systems<R: Resource + Clone>(app: &mut App) {
```
cBournhonesque, line 203: maybe we also need an observer that adds the `ResourceHistory` resource when the resource first gets added

cBournhonesque: oh i see you insert it directly in the protocol function, I think that works too!

cBournhonesque:
This looks really good overall! I just left a couple comments

lightyear/src/utils/ready_buffer.rs
nick-e: According to the docs, `std::collections::BinaryHeap` is a max-heap: https://doc.rust-lang.org/std/collections/struct.BinaryHeap.html#method.peek

I've added a unit test for the new function. Based on the other functions' comments it looks some of them assume the buffer is a min-heap.

lightyear/src/utils/ready_buffer.rs
nick-e: `ItemWithReadyKey` implements `PartialEq` so that it can implement `Ord` so that it can be used in a `BinaryHeap` within `ReadyBuffer`. `ReadyBuffer`'s comments describe its functions as dependent on the order of the key (e.g. a tick or instant) and does not mention being dependent on the order of the item's value.

`ReadyBuffer` doesn't implement `PartialEq` so `ready_buffer_1 == ready_buffer_2` doesn't compile but if it did then the equality would describe whether or not the two buffers have the same number of keys.

lightyear/src/utils/ready_buffer.rs
cBournhonesque: Yes but I inverted the `Ord` order of `ItemWithReadyKey` to make it a min-heap, or I thought..
I'll have to take a closer look

cBournhonesque:
Thanks! the changes look great

cBournhonesque:
Coming back to this, a `check_rollback<R>` function is probably needed as well?

PR #623, opened 2024-09-04
Rollback time resource during rollback

nick-e:
Ensure that the `Fixed<Time>` resource is accurate during a rollback by rollbacking the time resource when a rollback starts and advancing the time after each rollback tick. This is a continuation of https://github.com/cBournhonesque/lightyear/pull/619. This PR takes into account `Time<Fixed>`'s overstep.

I originally thought that using `add_resource_rollback::<Time<Fixed>>()` and running the `RunFixedMainLoop` schedule would be the solution but `RunFixedMainLoop` requires that the real time clock also progress in order to correctly advance the `Time<Fixed>` resource.

Fixes https://github.com/cBournhonesque/lightyear/issues/618

### DIFF
--- a/lightyear/src/client/prediction/rollback.rs
+++ b/lightyear/src/client/prediction/rollback.rs
@@ -876,6 +876,8 @@ mod unit_tests {
 /// More general integration tests for rollback
 #[cfg(test)]
 mod integration_tests {
+    use std::time::Duration;
+
     use super::test_utils::*;
 
     use crate::prelude::client::*;
@@ -883,23 +885,25 @@ mod integration_tests {
     use crate::tests::stepper::BevyStepper;
     use bevy::prelude::*;
 
-    fn increment_component(
-        mut commands: Commands,
-        mut query_networked: Query<(Entity, &mut ComponentSyncModeFull), With<Predicted>>,
-    ) {
-        for (entity, mut component) in query_networked.iter_mut() {
-            component.0 += 1.0;
-            if component.0 == 5.0 {
-                commands.entity(entity).remove::<ComponentSyncModeFull>();
+    fn setup(increment_component: bool) -> (BevyStepper, Entity, Entity) {
+        fn increment_component_system(
+            mut commands: Commands,
+            mut query_networked: Query<(Entity, &mut ComponentSyncModeFull), With<Predicted>>,
+        ) {
+            for (entity, mut component) in query_networked.iter_mut() {
+                component.0 += 1.0;
+                if component.0 == 5.0 {
+                    commands.entity(entity).remove::<ComponentSyncModeFull>();
+                }
             }
         }
-    }
 
-    fn setup() -> (BevyStepper, Entity, Entity) {
         let mut stepper = BevyStepper::default();
-        stepper
-            .client_app
-            .add_systems(FixedUpdate, increment_component);
+        if increment_component {
+            stepper
+                .client_app
+                .add_systems(FixedUpdate, increment_component_system);
+        }
         // add predicted/confirmed entities
         let tick = stepper.client_tick();
         let confirmed = stepper
@@ -929,6 +933,102 @@ mod integration_tests {
     }
 
     /// Test that:
+    /// - the `Time` resource's elapsed is rollbacked to the first tick of the rollback
+    /// - the `Time` resource's elapsed time is advanced correctly during the rollback
+    /// - the `Time` resource's delta during a rollback is the `Time<Fixed>`'s delta
+    #[test]
+    fn test_rollback_time_resource() {
+        #[derive(Debug, PartialEq)]
+        struct TimeSnapshot {
+            is_rollback: bool,
+            delta: Duration,
+            elapsed: Duration,
+        }
+
+        #[derive(Resource, Default, Debug)]
+        struct TimeTracker {
+            snapshots: Vec<TimeSnapshot>,
+        }
+
+        // Record the time resource's values for each tick.
+        fn track_time(
+            time: Res<Time>,
+            mut time_tracker: ResMut<TimeTracker>,
+            rollback: Res<Rollback>,
+        ) {
+            time_tracker.snapshots.push(TimeSnapshot {
+                is_rollback: rollback.is_rollback(),
+                delta: time.delta(),
+                elapsed: time.elapsed(),
+            });
+        }
+
+        let (mut stepper, confirmed, predicted) = setup(false);
+
+        // Insert arbitrary predicted component into confirmed. Needed to
+        // trigger a rollback.
+        stepper
+            .client_app
+            .world_mut()
+            .entity_mut(confirmed)
+            .insert(ComponentSyncModeFull(0.0));
+        stepper.frame_step();
+
+        // Check that the component got synced.
+        assert_eq!(
+            stepper
+                .client_app
+                .world()
+                .get::<ComponentSyncModeFull>(predicted)
+                .unwrap(),
+            &ComponentSyncModeFull(0.0)
+        );
+
+        // Trigger 2 rollback ticks by changing the confirmed's predicted
+        // component's value and setting the confirmed's tick to two ticks ago.
+        let tick = stepper.client_tick();
+        stepper
+            .client_app
+            .world_mut()
+            .get_mut::<ComponentSyncModeFull>(confirmed)
+            .unwrap()
+            .0 = 1.0;
+        received_confirmed_update(&mut stepper, confirmed, tick - 2);
+        stepper.client_app.insert_resource(TimeTracker::default());
+        stepper.client_app.add_systems(FixedUpdate, track_time);
+
+        let time_before_next_tick = *stepper.client_app.world().resource::<Time<Fixed>>();
+
+        stepper.frame_step();
+
+        // Verify that the 2 rollback ticks and regular tick occurred with the
+        // correct delta times and elapsed times.
+        let time_tracker = stepper.client_app.world().resource::<TimeTracker>();
+        assert_eq!(
+            time_tracker.snapshots,
+            vec![
+                TimeSnapshot {
+                    is_rollback: true,
+                    delta: stepper.tick_duration,
+                    elapsed: time_before_next_tick.elapsed() - stepper.tick_duration
+                },
+                TimeSnapshot {
+                    is_rollback: true,
+                    delta: stepper.tick_duration,
+                    elapsed: time_before_next_tick.elapsed()
+                },
+                TimeSnapshot {
+                    is_rollback: false,
+                    delta: stepper.tick_duration,
+                    elapsed: time_before_next_tick.elapsed() + stepper.tick_duration
+                }
+            ]
+        );
+
+        println!("{:?}", stepper.client_app.world().resource::<TimeTracker>());
+    }
+
+    /// Test that:
     /// - we remove a component from the predicted entity
     /// - rolling back before the remove should re-add it
     ///   We are still able to rollback properly (the rollback adds the component to the predicted entity)
@@ -937,7 +1037,7 @@ mod integration_tests {
         // tracing_subscriber::FmtSubscriber::builder()
         //     .with_max_level(tracing::Level::INFO)
         //     .init();
-        let (mut stepper, confirmed, predicted) = setup();
+        let (mut stepper, confirmed, predicted) = setup(true);
         // insert component on confirmed
         stepper
             .client_app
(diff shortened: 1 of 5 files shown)


nick-e:
I tested the fix using the replication process found in https://github.com/cBournhonesque/lightyear/issues/618

lightyear/src/protocol/component.rs
```diff
@@ -1108,10 +1107,10 @@ impl AppComponentExt for App {
         }
     }
 
+    /// Do not use `Time<Fixed>` for `R`. `Time<Fixed>` is already rollbacked.
```
cBournhonesque, line 1110: This comment probably shouldn't be there, right? This is for users who want to use `add_resource_rollback`

cBournhonesque:
Sorry I didn't have time to review in detail, but apparently just using the replication rollback to rollback Time<Fixed> and Time<Virtual> was not enough? We still need to have some ad-hoc logic?

> I tested the fix using the replication process found in https://github.com/cBournhonesque/lightyear/issues/618

What were the results? There's no rollbacks anymore?

oxkitsune:
This noticeably improved FPS drops I was experiencing during rollback. However there are still rollbacks for me, although those might be related to other things

nick-e:
I updated the PR to correctly advance the elapsed time of the fixed time resource during a rollback.

We can't use `add_resource_rollback::<Time<Fixed>>::()` because we lose `Time<Fixed>`'s **current** overstep when the resource rollbacks. We don't care what its overstep was when rollbacked. Overstep is used to determine when FixedMain should tick based on real time. Rollback ticks aren't run in real time.

On top of that, the only way to progress `Time<Fixed>` with accurate overstep is by running `RunFixedMainLoop` because overstep is private. The issue with that is the `RunFixedMainLoop` schedule progresses `Time<Fixed>` using `Time<Virtual>` which relies on real time. We would have to keep track of the real time for every tick that needs to be rollbacked, add them up, and apply it to `Time<Virtual>` before running `RunFixedMainLoop`. It gets more complicated because `Time<Virtual>` has `max_delta` which caps how much time can change and some rollbacks may be farther behind than that.

As a rule of thumb, `add_resource_rollback()` should only be used for resources that will **only** be modified by systems run in the `FixedMain` schedule. This is because rollbacks only run the `FixedMain` schedule. `Time<Fixed>` is modified by `bevy_time::fixed::run_fixed_main_schedule()` which is run outside of the `FixedMain` schedule. I think it's worth mentioning so I added it into the `add_resource_rollback()` doc comments.

# Testing
I tested this PR by adding the following to `client::handle_character_actions()` with the `avian_3d_character` example:
```rust
if let Some(ref rb) = rollback {
    if let Some(tick) = rb.get_rollback_tick() {
        info!(
            "    rb-{:?}, delta: {}, elapsed: {}",
            tick,
            time.delta_seconds(),
            time.elapsed_seconds()
        );
    } else {
        info!(
            "not-rb-{:?}, delta: {}, elapsed: {}",
            tick_manager.tick(),
            time.delta_seconds(),
            time.elapsed_seconds()
        );
    }
} else {
    info!(
        "{:?}, delta: {}, elapsed: {}",
        tick_manager.tick(),
        time.delta_seconds(),
        time.elapsed_seconds()
    );
}
```
This function is only run within a system within the `FixedMain` schedule.

I set the latency to 100ms and a packet loss of 0.1 in order to incur multi-tick rollbacks. Here are some logs of the `avian_3d_character` client run with `cargo run --release -- client -c 1` without the PR:
```
2024-09-05T16:03:41.208296Z  INFO avian_3d_character::client: not-rb-Tick(257), delta: 0.015625, elapsed: 3.796875
2024-09-05T16:03:41.224970Z  INFO avian_3d_character::client: not-rb-Tick(258), delta: 0.015625, elapsed: 3.8125
2024-09-05T16:03:41.237516Z  INFO avian_3d_character::client: not-rb-Tick(259), delta: 0.015625, elapsed: 3.828125
2024-09-05T16:03:41.254084Z  INFO avian_3d_character::client: not-rb-Tick(260), delta: 0.015625, elapsed: 3.84375
2024-09-05T16:03:41.270763Z  INFO avian_3d_character::client: not-rb-Tick(261), delta: 0.015625, elapsed: 3.859375
2024-09-05T16:03:41.287490Z  INFO avian_3d_character::client: not-rb-Tick(262), delta: 0.015625, elapsed: 3.875
2024-09-05T16:03:41.299996Z  INFO avian_3d_character::client: not-rb-Tick(263), delta: 0.015625, elapsed: 3.890625
2024-09-05T16:03:41.312512Z  INFO avian_3d_character::client:     rb-Tick(257), delta: 0.004247744, elapsed: 3.9034278 <-- wrong delta; elapsed not rollbacked
2024-09-05T16:03:41.313247Z  INFO avian_3d_character::client:     rb-Tick(258), delta: 0.004247744, elapsed: 3.9034278 <-- elapsed not advancing
2024-09-05T16:03:41.313825Z  INFO avian_3d_character::client:     rb-Tick(259), delta: 0.004247744, elapsed: 3.9034278
2024-09-05T16:03:41.314345Z  INFO avian_3d_character::client:     rb-Tick(260), delta: 0.004247744, elapsed: 3.9034278
2024-09-05T16:03:41.314856Z  INFO avian_3d_character::client:     rb-Tick(261), delta: 0.004247744, elapsed: 3.9034278
2024-09-05T16:03:41.315427Z  INFO avian_3d_character::client:     rb-Tick(262), delta: 0.004247744, elapsed: 3.9034278
2024-09-05T16:03:41.315992Z  INFO avian_3d_character::client:     rb-Tick(263), delta: 0.004247744, elapsed: 3.9034278
2024-09-05T16:03:41.318197Z  INFO avian_3d_character::client: not-rb-Tick(264), delta: 0.015625, elapsed: 3.90625
```

Here are some logs with the PR:
```
2024-09-05T16:00:26.780969Z  INFO avian_3d_character::client: not-rb-Tick(129), delta: 0.015625, elapsed: 1.828125
2024-09-05T16:00:26.792864Z  INFO avian_3d_character::client: not-rb-Tick(130), delta: 0.015625, elapsed: 1.84375
2024-09-05T16:00:26.809763Z  INFO avian_3d_character::client: not-rb-Tick(131), delta: 0.015625, elapsed: 1.859375
2024-09-05T16:00:26.826218Z  INFO avian_3d_character::client: not-rb-Tick(132), delta: 0.015625, elapsed: 1.875
2024-09-05T16:00:26.838714Z  INFO avian_3d_character::client: not-rb-Tick(133), delta: 0.015625, elapsed: 1.890625
2024-09-05T16:00:26.855291Z  INFO avian_3d_character::client: not-rb-Tick(134), delta: 0.015625, elapsed: 1.90625
2024-09-05T16:00:26.871923Z  INFO avian_3d_character::client: not-rb-Tick(135), delta: 0.015625, elapsed: 1.921875
2024-09-05T16:00:26.880931Z  INFO avian_3d_character::client:     rb-Tick(129), delta: 0.015625, elapsed: 1.828125 <-- correct delta; elapsed rollbacked
2024-09-05T16:00:26.881672Z  INFO avian_3d_character::client:     rb-Tick(130), delta: 0.015625, elapsed: 1.84375  <-- elapsed advancing correctly
2024-09-05T16:00:26.882308Z  INFO avian_3d_character::client:     rb-Tick(131), delta: 0.015625, elapsed: 1.859375
2024-09-05T16:00:26.882863Z  INFO avian_3d_character::client:     rb-Tick(132), delta: 0.015625, elapsed: 1.875
2024-09-05T16:00:26.883419Z  INFO avian_3d_character::client:     rb-Tick(133), delta: 0.015625, elapsed: 1.890625
2024-09-05T16:00:26.883931Z  INFO avian_3d_character::client:     rb-Tick(134), delta: 0.015625, elapsed: 1.90625
2024-09-05T16:00:26.884464Z  INFO avian_3d_character::client:     rb-Tick(135), delta: 0.015625, elapsed: 1.921875
2024-09-05T16:00:26.888542Z  INFO avian_3d_character::client: not-rb-Tick(136), delta: 0.015625, elapsed: 1.9375
```

There are still constant rollbacks. Still looking into that.

lightyear/src/protocol/component.rs
nick-e: Updated the doc comments to add clarity. Let me know what you think!

nick-e:
I added a test for verifying that the time resource is accurate during a rollback. I can confirm that it fails without this PR.

lightyear/src/client/prediction/rollback.rs
```diff
@@ -568,6 +569,38 @@ pub(crate) fn prepare_rollback_resource<R: Resource + Clone>(
...
+    }
+
+    // Difference between the current time and the time of the first tick of
+    // the rollback.
+    let rollback_time_offset = (num_rollback_ticks - 1) as u32 * rollback_fixed_time.timestep();
+
+    let rollback_elapsed_time = current_fixed_time.elapsed() - rollback_time_offset;
+    rollback_fixed_time.advance_to(rollback_elapsed_time - rollback_fixed_time.timestep());
+    // This sets the fixed time's delta to the correct amount.
+    rollback_fixed_time.advance_by(rollback_fixed_time.timestep());
```
cBournhonesque, line 599: I think this could have a clarifying comment.
`Time<Fixed>.delta` is set to the value provided in `advance_by` (or `advance_to`), so we want to call `advance_by(rollback_fixed_time.timestep())` at the end to set the delta value that is expected.

(I had to read the `advance_by` source to check this, but it would be good to mention it here)

lightyear/src/client/prediction/rollback.rs
```diff
@@ -590,13 +623,44 @@ pub(crate) fn run_rollback(world: &mut World) {
...
+
+    // Run the fixed update schedule (which should contain ALL
+    // predicted/rollback components and resources). This is similar to what
+    // `bevy_time::fixed::run_fixed_main_schedule()` does
     for i in 0..num_rollback_ticks {
         debug!("Rollback tick: {:?}", current_rollback_tick + i);
+
+        // Set the rollback tick's generic time resource to the fixed time
+        // resource that was just advanced.
+        *world.resource_mut::<Time>() = world.resource::<Time<Fixed>>().as_generic();
```
cBournhonesque, line 643: Why do we want to also rollback the non-fixed-time Time?

lightyear/src/shared/time_manager.rs
```diff
@@ -56,7 +56,7 @@ pub struct TimeManager {
     /// The relative speed set by the client.
     pub base_relative_speed: f32,
     /// Should we speedup or slowdown the simulation to sync the ticks?
-    /// >1.0 = speedup, <1.0 = slowdown
+    /// \>1.0 = speedup, <1.0 = slowdown
```
cBournhonesque, line 59: Is this a typo?

cBournhonesque:
Thanks for the explanation on why rollbacking Time<Fixed> + running RunFixedMainLoop wouldn't work.
I also appreciate the tests and the logs!

I'm convinced by this PR, making sure that Time<Fixed> was handled correctly during rollback was something I had completely overlooked!

I left a couple comments but otherwise I think it's ready to merge. Thanks for the awesome work!

cBournhonesque (approved):

lightyear/src/client/prediction/rollback.rs
nick-e: `Time` is not a non-fixed time but instead a time without a "context". It could represent fixed time or some other kind of time like `Time<Virtual>`. It's my understanding that systems within the `FixedMain` should be safe to assume that the value of the generic `Time` resource it retrieves via `Res<Time>` represents the `Time<Fixed>` resource.

lightyear/src/shared/time_manager.rs
nick-e: `/// >` is a table entry just like `/// *` is a bullet point. The rust doc comment has to escape the `>` to denote that the character is meant to represent the greater-than sign and not the start of a table entry. The linter was failing due to that issue.

lightyear/src/client/prediction/rollback.rs
nick-e: Updated!

lightyear/src/shared/time_manager.rs
nick-e: https://rust-lang.github.io/rust-clippy/master/index.html#/doc_lazy_continuation

cBournhonesque:
@nick-e Could a reason for the avian rollbacks be that avian is using `Time<Physics>` and `Time<Substeps>` which should be rollbacked in a similar manner to `Time<Fixed>`: https://github.com/Jondolf/avian/blob/v0.2.0/src/schedule/mod.rs#L252 ?

PR #624, opened 2024-09-05
Spaceships example: change Without<Confirmed> to With<Predicted> in Visual Interp setup

RJ:
I noticed replication on the client adds Position before it adds Confirmed, causing the Confirmed entities to also get the visual interpolation components. 

I changed the filters to make more sense, and not bothering with interp on the server.

### DIFF
--- a/examples/spaceships/src/renderer.rs
+++ b/examples/spaceships/src/renderer.rs
@@ -79,9 +79,12 @@ impl Plugin for SpaceshipsRendererPlugin {
 // Non-wall entities get some visual interpolation by adding the lightyear
 // VisualInterpolateStatus component
 //
-// We query Without<Confirmed> instead of With<Predicted> so that the server's gui will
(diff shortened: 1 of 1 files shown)


cBournhonesque:
Hm that's strange; or is it because of pre-spawning?

RJ:
on main at the moment, if you open the egui inspector, you can see that as well as the players, all the balls have the visual interp components on their Confirmed entities as well as on the Predicted. so it's nothing to do with predictive spawning – all those entities are server authoritative anyway.

only explanation i can think of it that Position and Rotation are inserted before the Confirmed component, causing the `Without<Confirmed>` filter to be useless.

RJ:
from a quick look, perhaps because ShouldBePredicted is used to trigger spawning the Predicted entity, and only then is the Confirmed component added to the confirmed entity.. not really a bug, just caught me out.

PR #626, opened 2024-09-06
Add book section describing component prediction criteria

nick-e:
Add a section to the lightyear book discussing which components should be predicted if a game wants to use prediction. It took me some time to understand which components should be predicted so hopefully this helps others! No worries if you don't want to add it!

### DIFF
--- a/book/src/concepts/advanced_replication/prediction.md
+++ b/book/src/concepts/advanced_replication/prediction.md
@@ -42,6 +42,40 @@ Then on the server, you can replicate back the entity to the original client. Th
 instead of spawning a new separate `Predicted` entity,
 it will re-use the existing entity that had the `PrePredicted` component!
 
+## Which components should I predict?
+
+You want to predict components for entities that will live in the Predicted timeline, i.e. the timeline that will see changes immediately based on client inputs. However it doesn't mean that every component needs to be actively be predicted with `ComponentSyncMode::Full`, i.e. with rollbacks enabled. A lot of components can be computed from other components and don't need to be predicted or even sent through the network. Here is a quick explanation of which components should be predicted.
+
+As a rule of thumb, a component should be predicted with `ComponentSyncMode::Full` if it meets the following criteria:
+
+1. Cannot be calculated using the **predicted** components available within the **current** tick.
+2. May be modified after creation.
+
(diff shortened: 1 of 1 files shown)


book/src/concepts/advanced_replication/prediction.md
```diff
@@ -42,6 +42,38 @@ Then on the server, you can replicate back the entity to the original client. Th
...
 it will re-use the existing entity that had the `PrePredicted` component!
 
+## Which components should I predict?
+
+As a rule of thumb, if you want to enable prediction for your game then a component should be predicted if it meets the following criteria:
+
+1. Cannot be calculated using the **predicted** components available within the **current** tick.
+2. May be modified after creation.
+
+For example, let's use the `avian3d::position::Position` component provided by the physics simulator `avian`. It describes the position of a 3D object. It's value is modified at the end of each tick by the `avian` physics simulator so it meets the second criteria. If a system wants to know the value of a `Position` component during a given tick, it has no way of calculating that value. Instead, the system will query the `Position` component whose value was calculated in the **previous** tick by the `avian` physics simulator. This means that `Position` also meets the first criteria and so it should be predicted.
```
cBournhonesque, line 52: typo: `Its value is modified`

book/src/concepts/advanced_replication/prediction.md
```diff
@@ -42,6 +42,38 @@ Then on the server, you can replicate back the entity to the original client. Th
 instead of spawning a new separate `Predicted` entity,
 it will re-use the existing entity that had the `PrePredicted` component!
 
+## Which components should I predict?
```
cBournhonesque, line 45: I think it would be helpful to add a short intro here:

you want to predict components for entities that will live in the `Predicted` timeline, i.e. the timeline that will see changes immediately based on client inputs. However it doesn't mean that every component needs to be actively be predicted with `ComponentSyncMode::Full`, i.e. with rollbacks enabled. A lot of components can be computed from other components and don't need to be predicted or even sent through the network. Here is a quick explanation of which components should be predicted

cBournhonesque:
Every doc/book improvement is greatly appreciated!
You probably understand lightyear better than 95% of users, so if you had some misunderstandings it's probable that others might be confused as well

PR #628, opened 2024-09-07
Add rollback event for debugging rollback causes

mbrea-c:
Currently there's no easily accessible way of checking what components are causing rollbacks. 

With this PR, we fire an event whenever a rollback check detects a mismatch which listeners can then aggregate as needed to get statistics on rollback causes.

### DIFF
--- a/lightyear/src/client/prediction/rollback.rs
+++ b/lightyear/src/client/prediction/rollback.rs
@@ -1,3 +1,4 @@
+use std::any::TypeId;
 use std::fmt::Debug;
 use std::ops::{Deref, DerefMut};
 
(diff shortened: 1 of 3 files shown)


lightyear/src/client/prediction/rollback.rs
```diff
@@ -113,6 +125,7 @@ pub(crate) fn check_rollback<C: SyncComponent>(
     // We use Option<> because the predicted component could have been removed while it still exists in Confirmed
     confirmed_query: Query<(Entity, Option<&C>, Ref<Confirmed>)>,
     rollback: Res<Rollback>,
+    mut evw_rollback: EventWriter<RollbackEvent>,
```
cBournhonesque, line 128: This would prevent the `check_rollback` systems for multiple components from running in parallel, so I'm not sure it's something we want to have currently unless we redesign those systems to have only one system for all predicted components

mbrea-c: Hmm, good point. How about if `RollbackEvent` is generic on the component type? It'll be a bit more work for users to aggregate but it won't prevent paralellism

PR #636, opened 2024-09-09
Add missing `reflect(Component)`s

philpax:
Several key already-reflectable components didn't have `#[reflect(Component)]` attached, which means that their `ReflectComponent` couldn't be retrieved. I've gone through and added this where appropriate to components that already had `Reflect` derived.


cBournhonesque (approved):
Thanks!

Issue #639, opened 2024-09-12
`spawn_interpolated_entity` panics when authority is transferred away from a client-spawned entity


philpax:
When transferring authority of an interpolated entity away from that client, the following code will panic (determined through a local fork that replaced the `unwrap`s with `expect`s):

https://github.com/cBournhonesque/lightyear/blob/3a610d9d808d3166d51c47ecef2bd381760aa973/lightyear/src/client/interpolation/spawn.rs#L37-L40

as it appears that `get_confirmed_tick` returns a `None` in this case.

To reproduce:
1. Join the server with two clients, C1 and C2.
2. Spawn an entity on C1 and replicate it to the server using the default `Replicate` bundle.
3. On the server, replicate it to other peers using the following replicate bundle:
    ```rust
    Replicate {
        target: ReplicationTarget {
            target: NetworkTarget::All,
        },
        authority: AuthorityPeer::Client(client_id),
        sync: SyncTarget {
            prediction: NetworkTarget::None,
            interpolation: NetworkTarget::All,
        },
        controlled_by: ControlledBy {
            target: NetworkTarget::None,
            lifetime: Lifetime::Persistent,
        },
        group: ReplicationGroup::new_from_entity(),
        ..default()
    }
    ```
4. Transfer authority of that entity to C2. C1 will now panic in the above code.

cBournhonesque:
I took another look at this.
I think I fixed some of the underlying issues related to the entity mapping.
The problem I have now is:
- when the entity gets transferred to C2, how do we make sure that server sends a ShouldBeInterpolated to C1?

This doesn't happen currently because:
- sending `ShouldBeInterpolated` and `ShouldBePredicted` is only added when we send an entity-spawn; in this case there is no real entity-spawn. Even if I updated the code to still send a fake entity-spawn (with `ShouldBeInterpolated`) when the authority changes, then C1 would just ignore the message because the entity already exists

I see several options:
- instead of the server sending `ShouldBeInterpolated` and `ShouldBePredicted`, the client should run an observer to decide if they want to add Prediction/Interpolation. Then in the example above the client would directly spawn Predicted/Confirmed/Interpolated. While the client has authority, only Confirmed is used (or maybe Predicted). But when the client loses authority, the client would switch to the `Interpolated` entity (or spawn a new one if needed).

Maybe we could provide a component `SyncBehaviour` that users could add on client entities (on creation or when it's created via replication) to specify if interpolation should be done etc.

[...]

PR #642, opened 2024-09-12
Fix 639

cBournhonesque:
Related to  https://github.com/cBournhonesque/lightyear/issues/639
The problem is that we used to only do entity mapping from the Sender -> Receiver, i.e. the entity map was **only** present on the receiver. With authority transfer we updated entity mapping, and it can be done both on the sending side or on the receiving side.

Some of our existing logic on the receiver (for example: https://github.com/cBournhonesque/lightyear/blob/1e1cbfadc6fcdcfca61a7b6cf091e24ceaf4a0ba/lightyear/src/shared/replication/receive.rs#L196) still work under the expectation that the entity map only exists on the receiver. (and that **every local entity** is present in the entity map. This is not the case anymore, which is why this function fails (https://github.com/cBournhonesque/lightyear/blob/1e1cbfadc6fcdcfca61a7b6cf091e24ceaf4a0ba/lightyear/src/client/interpolation/spawn.rs#L39).
https://github.com/cBournhonesque/lightyear/blob/1e1cbfadc6fcdcfca61a7b6cf091e24ceaf4a0ba/lightyear/src/shared/replication/receive.rs#L177

Potential solutions:
- include the received tick in `Replicated` so that ALL entities have it (not just the ones with prediction/interp). Could work but maybe inefficient?
- In general using the ChannelReceiver to get the confirmed tick seems very 'flimsy'. Maybe use something else?
- Fix the code paths that still expect all entities to be present in the entity map

This PR currently has a unit test but no fix

### DIFF
--- a/examples/lobby/src/server.rs
+++ b/examples/lobby/src/server.rs
@@ -56,7 +56,6 @@ impl Plugin for ExampleServerPlugin {
     }
 }
 
-
 /// System to start the dedicated server at Startup
 fn start_dedicated_server(mut commands: Commands) {
     commands.replicate_resource::<Lobbies, Channel1>(NetworkTarget::All);
@@ -238,8 +237,7 @@ mod lobby {
                         host: lobby.host,
                     },
                 );
-            }
-            else {
+            } else {
                 if host.is_none() {
                     // one of the players asked for the game to start
                     for player in &lobby.players {
--- a/lightyear/src/client/input/leafwing.rs
+++ b/lightyear/src/client/input/leafwing.rs
@@ -375,6 +375,8 @@ fn get_delayed_action_state<A: LeafwingUserAction>(
         //  the problem is that the Leafwing Plugin works on ActionState directly...
         if let Some(delayed_action_state) = input_buffer.get(delayed_tick) {
             *action_state = delayed_action_state.clone();
+            // dbg!(input_buffer);
+            // dbg!(delayed_tick);
             debug!(
                 ?entity,
                 ?delayed_tick,
@@ -383,6 +385,7 @@ fn get_delayed_action_state<A: LeafwingUserAction>(
                 input_buffer
             );
         }
+        // TODO: if we don't find an ActionState in the buffer, should we reset the delayed one to default?
     }
     // if let Some(mut action_state) = global_action_state {
     //     *action_state = global_input_buffer.get_last().unwrap().clone();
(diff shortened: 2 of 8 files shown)


cBournhonesque:
I'm merging this because it provides some useful changes:
- change some data that depends on `remote_entities` to instead depend on `local_entities` -> this is important because in some cases we might not know about the remote entities.

- I tried to implement a change where when you change replication from c1 to s or c2, then we potentially send a `ShouldBeInterpolated` or `ShouldBePredicted` to c1. Unfortunately it doesn't work because c1 doesn't accept it since it hasn't received yet the message that it doesn't have authority anymore.. In general this was a bit of a hack, so i'll keep thinking of better ways to handle Prediction/Interpolation.

@philpax in the meantime, is it possible for you to always start spawning entities on the server, instead of on clients? That would make things simpler, and I suspect that interpolation/prediction won't panic

philpax:
I haven't been able to work on the project that I originally needed this for, so I can't test this change exactly. 

However, I've been working on another project using Lightyear, also using client replication, and I've had to revert this change on my local branch because I'm being spammed with log errors that look like this:
```
2024-10-02T21:53:19.660639Z ERROR lightyear::shared::replication::receive: updating confirmed tick for entity remote_tick=Tick(8305) local_entity=Entity { index: 112, generation: 1 }
2024-10-02T21:53:19.677335Z ERROR lightyear::shared::replication::receive: updating confirmed tick for entity remote_tick=Tick(8307) local_entity=Entity { index: 112, generation: 1 }
2024-10-02T21:53:19.760975Z ERROR lightyear::shared::replication::receive: updating confirmed tick for entity remote_tick=Tick(8315) local_entity=Entity { index: 112, generation: 1 }
2024-10-02T21:53:19.777760Z ERROR lightyear::shared::replication::receive: updating confirmed tick for entity remote_tick=Tick(8317) local_entity=Entity { index: 112, generation: 1 }
```

No interpolation or prediction, just replicating a single entity from the client. The errors occur on both the server and receiving clients.

> @philpax in the meantime, is it possible for you to always start spawning entities on the server, instead of on clients? That would make things simpler, and I suspect that interpolation/prediction won't panic

This is a bit difficult for the project that this was originally for, but not impossible. For now, I'm content operating without interaction/prediction, but I'll revisit that at a later stage.

cBournhonesque:
Ah sorry about that; I just removed the extra logs (in `main`)

PR #647, opened 2024-09-17
Map entities when rollbacking predicted component

nick-e:
Map the entities within a predicted component when it is rollbacked. If they are not mapped then the predicted component's entities will be the same as its confirmed component counterpart when a rollback starts which may not be desired.

### DIFF
--- a/lightyear/src/client/prediction/rollback.rs
+++ b/lightyear/src/client/prediction/rollback.rs
@@ -226,7 +226,6 @@ pub(crate) fn prepare_rollback<C: SyncComponent>(
     // We use Option<> because the predicted component could have been removed while it still exists in Confirmed
     mut predicted_query: Query<
         (
-            Entity,
             Option<&mut C>,
             &mut PredictionHistory<C>,
             Option<&mut Correction<C>>,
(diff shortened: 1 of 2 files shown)


cBournhonesque:
Ah, so the problem is that I already perform conversion from Confirmed to Predicted when inserting a component on the Predicted entity or in the `PredictedHistory`; but I wasn't doing the conversion during rollbacks?

Do you think you could add a unit test that would show that the current code is failing?

nick-e:
Yep! I can also work on a unit test

nick-e:
Added a test that I can confirm fails without this PR.

lightyear/src/client/prediction/rollback.rs
```diff
@@ -932,6 +943,109 @@ mod integration_tests {
         (stepper, confirmed, predicted)
     }
 
+    /// Test that the entities of a predicted component marked as to be
+    /// entity-mapped are mapped when rollbacked.
+    #[test]
+    fn test_rollback_entity_mapping() {
+        #[derive(Component, Serialize, Deserialize, Clone, Copy, PartialEq)]
+        struct ComponentWithEntity(Entity);
```
cBournhonesque, line 951: I already provide a `ComponentMapEntities` in shared/tests/protocol if you want to skip this setup code

lightyear/src/client/prediction/rollback.rs
```diff
@@ -932,6 +943,109 @@ mod integration_tests {
...
+            .register_component::<ComponentWithEntity>(ChannelDirection::Bidirectional)
+            .add_prediction(ComponentSyncMode::Full)
+            .add_map_entities();
+        stepper
+            .server_app
+            .register_component::<ComponentWithEntity>(ChannelDirection::Bidirectional)
+            .add_prediction(ComponentSyncMode::Full)
+            .add_map_entities();
+        let tick = stepper.client_tick();
+        let confirmed = stepper
```
cBournhonesque, line 971: Maybe this whole predicted/confirmed setup is easier to do by just replicating a server entity to the client? not sure

cBournhonesque (approved):
Looks good to me, thanks! I left a few nit comments on the test

lightyear/src/client/prediction/rollback.rs
nick-e: `ComponentMapEntities` uses simple prediction. Trying to override its prediction mode to full doesn't seem to work.

PR #653, opened 2024-09-30
Simplify wrapping_diff code

ambiso:
`wrapping_diff` can be simplified to `b.wrapping_sub(a) as i16`

I wrote a test to compare the two implementations so a reviewer can convince themselves that they are identical.

The test can be run with:

```sh
cargo test --release -- test_wrapping_diff_equivalence --ignored
```

`--release` is required because it tests all $2^{32}$ inputs.

The wrapping_diff_old can be removed before merge.

### DIFF
--- a/lightyear/src/utils/wrapping_id.rs
+++ b/lightyear/src/utils/wrapping_id.rs
@@ -155,54 +155,11 @@ pub fn wrapping_diff(a: u16, b: u16) -> i16 {
     b.wrapping_sub(a) as i16
 }
 
-// TODO: Remove
-fn wrapping_diff_old(a: u16, b: u16) -> i16 {
-    const MAX: i32 = i16::MAX as i32;
-    const MIN: i32 = i16::MIN as i32;
-    const ADJUST: i32 = (u16::MAX as i32) + 1;
-
-    let a: i32 = i32::from(a);
-    let b: i32 = i32::from(b);
-
-    let mut result = b - a;
-    if (MIN..=MAX).contains(&result) {
-        result as i16
-    } else if b > a {
-        result = b - (a + ADJUST);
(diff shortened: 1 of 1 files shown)


lightyear/src/utils/wrapping_id.rs
```diff
@@ -182,6 +152,11 @@ pub(crate) use wrapping_id;
 /// assert_eq!(wrapping_diff(0, 32768), -32768);
 /// ```
 pub fn wrapping_diff(a: u16, b: u16) -> i16 {
+    b.wrapping_sub(a) as i16
+}
+
+// TODO: Remove
```
Nul-led, line 158: old impl needs to be removed as expressed in description

lightyear/src/utils/wrapping_id.rs
```diff
@@ -215,6 +190,20 @@ mod sequence_compare_tests {
 
     wrapping_id!(Id);
 
+    #[ignore]
```
Nul-led, line 193: needs to be removed before merge too

Nul-led (changes requested):
Ran the test, seems to work

lightyear/src/utils/wrapping_id.rs
ambiso: Done

lightyear/src/utils/wrapping_id.rs
ambiso: Test removed

Nul-led (approved):

Nul-led:
@cBournhonesque ready to merge :)

cBournhonesque:
Thanks that is much better! I don't know why I didn't go for this, I guess I must have had a good reason at the time

PR #654, opened 2024-10-01
Expose and document `SerializeFns`

philpax:
While trying to use `register_component_serialize_fns`, I noticed that `SerializeFns` was not exposed. I've fixed this and also added some basic comments to explain its use.


cBournhonesque (approved):
Thank you!

PR #661, opened 2024-10-06
Support LWIM 15.1 triple axis input

msvbg:
LWIM 15.1 added a new triple axis input type, which breaks Lightyear. There were no tests here, ~and I didn't add any~ I added some. I just repeated the logic that was already in place for the other axis types.


cBournhonesque (approved):

cBournhonesque:
Thank you!

PR #665, opened 2024-10-07
Require UserAction bound for InputPlugin

Aceeri:
Improves the error you get a bit if you didn't implement UserAction for your input correctly.

Before:
image

After:
image


cBournhonesque (approved):
In general I try to leave bounds as general as possible (I think best practice is to only add bounds on the impl, not on the struct itself), but this looks relatively harmless.

Let's try it; thanks!

PR #666, opened 2024-10-07
Avoid panic when getting EntityCommands in hierarchy

msvbg:
Lightyear seems to panic when despawning children with an error message like:

```
thread 'Compute Task Pool (6)' panicked at lightyear/src/shared/replication/hierarchy.rs:111:30:
Attempting to create an EntityCommands for entity Entity { index: 720, generation: 1 }, which doesn't exist.
```

This PR adds a test case and attempts a fix.


cBournhonesque (approved):
Thanks! I thought that calling `iter_descendants` was guaranteeing that all the children existed, but I guess not?

msvbg:
Yeah I'm also a little confused why actually, but I have seen quite a few similar panics from my own code so I think the answer is no.

PR #670, opened 2024-10-10
Disable default leafwing features

### LINKED ISSUE #669: Enabling the `leafwing` feature adds 100 extra dependencies
Here's a minimal example of `Cargo.toml` to reproduce the issue:
```
[package]
name = "test_deps"
version = "0.1.0"
edition = "2021"

[dependencies]
bevy = { version = "0.14.2", default-features = false }
leafwing-input-manager = { version = "0.15.1", default-features = false }
lightyear = { version = "0.17.1", default-features = false }
```

^ This is a minimal setup that is needed to implement a headless Bevy app. It brings in ~200 crates (which is without the 100 extra).

But if we enable the `leafwing` feature like that: `lightyear = { version = "0.17.1", default-features = false, features = ["leafwing"] }` - cargo will now have to compile 309 dependencies.

I think the issue is that `lightyear` brings in `leafwing-input-manager` with default features enabled. But I'm not sure if they are all actually needed.
The added crates include `bevy_render`, for example, which is not ideal for a headless app.

Is it possible to disable default features for `leafwing-input-manager`? I guess we still need to verify that `lightyear` doesn't indeed need them.

mvlabat:
Fixes #669. I didn't test this extensively, but the change didn't break my pet project, so I hope it works.

### DIFF
--- a/lightyear/Cargo.toml
+++ b/lightyear/Cargo.toml
@@ -64,7 +64,7 @@ thiserror = "1.0.50"
 seahash = "4.1.0"
 
 # input
-leafwing-input-manager = { version = "0.15", optional = true }
+leafwing-input-manager = { version = "0.15", default-features = false, optional = true }
 
 # physics
 avian2d = { version = "0.1.1", optional = true, default-features = false }


RJ:
i ran into this too - trying to build for wasm, and LWIM with default features is pulling in stuff i don't want relating to clipboard management on the web (unstable apis).

mvlabat:
image
image

The change saved about 10m of build time for me as well. (The jobs weren't using any cache for rust artifacts.)

cBournhonesque:
Cool, thanks!

PR #687, opened 2024-11-09
fix: Remove `ring` to fix wasm32 web builds

### LINKED ISSUE #684: WASM Examples wont run in browser
### Summary

I've encountered an issue while trying to run any example specifically for WASM (e.g., `simple_box`, `interest_management`) using `trunk serve`. I’ve tested both the `main` branch and the latest tagged version (`0.17.0`). While the examples compile successfully to WebAssembly, they display a blank white page when opened in either Firefox or Chrome.

### Steps to Reproduce

1. Clone the repository and checkout either the `main` branch or the `0.17.0` tag.
2. Run `trunk serve` to build and serve any example, such as `simple_box` or `interest_management`.
3. Open the served page (default: `http://127.0.0.1:8080`) in a browser.

### Expected Behavior

The example should load and display the expected content.

### Actual Behavior
(Using the ```interest_manegement``` example)

A blank white page appears in both Firefox and Chrome. Opening the browser console shows the following error:
```Uncaught TypeError: The specifier “env” was a bare specifier, but was not remapped to anything. Relative module specifiers must start with “./”, “../” or “/”.```
Error originates from:
```interest_management-b07c367cc454b4e.js:1:30 (http://127.0.0.1:8080/interest_management-b07c367cc454b4e.js)```
The highlighted line in the code is:

```javascript
import * as __wbg_star0 from 'env';
```

### Analysis
This error seems to occur because WebAssembly is attempting to import an "env" symbol that doesn't exist, possibly due to a missing or incorrectly configured crate dependency.

### Debugging Attempts
Following the suggestions in [this discussion](https://github.com/rustwasm/wasm-bindgen/discussions/3500#discussioncomment-6334669), I attempted to inspect the .wasm file using wasm2wat to identify the missing symbol/crate. However, running wasm2wat on the file failed, showing this message:
```Error: Aborted(). Build with -s ASSERTIONS=1 for more info.```
I tried building with additional flags as suggested:
```bash
RUSTFLAGS="--cfg=web_sys_unstable_apis -C debuginfo=2" trunk serve
```
This did not resolve the issue.

### Environment
 * Browser: Firefox and Chrome (latest versions)
 * OS: Ubuntu 24.04
 * Trunk Version: trunk 0.21.4
 * Rust Version: 1.82.0

Any guidance on resolving this would be greatly appreciated (Though I got a feeling that this might be a "Me issue").

cdata:
Fixes #684 

This is admittedly a blunt solution to the problem. For uncertain reasons, the `ring` dependency does not seem compatible with `wasm32-unknown-unknown`/`wasm-bindgen` builds. But: `ring` only seems to be used for this ~25 line utility function. So, the fix proposed here is to remove the `ring` dependency and inline the utility function. Submitted for your consideration!

### DIFF
--- a/lightyear/Cargo.toml
+++ b/lightyear/Cargo.toml
@@ -28,7 +28,6 @@ webtransport = [
   "dep:xwt-core",
   "dep:xwt-web",
   "dep:web-sys",
-  "dep:ring",
   "dep:wasm-bindgen-futures",
 ]
 leafwing = ["dep:leafwing-input-manager"]
@@ -147,7 +146,6 @@ zstd = { version = "0.13.1", optional = true }
 
 [target."cfg(target_family = \"wasm\")".dependencies]
 console_error_panic_hook = { version = "0.1.7" }
-ring = { version = "0.17.8", optional = true, default-features = false }
 web-sys = { version = "0.3", optional = true, features = [
   "Document",
   "WebTransport",
--- a/lightyear/src/transport/webtransport/client_wasm.rs
+++ b/lightyear/src/transport/webtransport/client_wasm.rs
@@ -18,6 +18,36 @@ use crate::transport::error::{Error, Result};
 use crate::transport::io::IoState;
 use crate::transport::{BoxedReceiver, BoxedSender, PacketReceiver, PacketSender, Transport, MTU};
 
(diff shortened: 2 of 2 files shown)


Nul-led (approved):
Lgtm

cBournhonesque:
awesome, thank you

PR #689, opened 2024-11-11
Fixed example floor not being rendered

### LINKED ISSUE #686: avian_3d_character example floor does not render in host-server mode
As title says, if example is run in host-server mode, the floor is not rendered. I narrowed it down to not having the Replicated component (instead it has Replicating) when run in host-server mode.

I am currently trying the lightyear for my project, and this is the exact issue I stumbled upon myself. I need static level objects to appear on both remote and local clients. What is the best way to achieve that?

sQu1rr:
The floor is not being rendered in the avian_3d_character example when run in host-server mode.

The issue was due to cosmetics being applied to the floor with a `Replicated` component, which can only be present on the client. The fix is to apply cosmetics to the floor unconditionally.

Fixes #686

### DIFF
--- a/examples/avian_3d_character/src/renderer.rs
+++ b/examples/avian_3d_character/src/renderer.rs
@@ -154,7 +154,7 @@ fn add_character_cosmetics(
 /// do not exist since floors aren't predicted.
 fn add_floor_cosmetics(
     mut commands: Commands,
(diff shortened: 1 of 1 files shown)


cBournhonesque:
Thanks!

PR #711, opened 2024-11-26
Compilation fixes

RJ:
fixing some compilation errors caused by updating deps

### DIFF
--- a/lightyear/src/transport/webtransport/server.rs
+++ b/lightyear/src/transport/webtransport/server.rs
@@ -58,7 +58,7 @@ impl ServerTransportBuilder for WebTransportServerSocketBuilder {
 
         let mut config = ServerConfig::builder()
             .with_bind_address(self.server_addr)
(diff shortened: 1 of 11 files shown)


lightyear/src/transport/webtransport/server.rs
```diff
@@ -58,7 +58,7 @@ impl ServerTransportBuilder for WebTransportServerSocketBuilder {
 
         let mut config = ServerConfig::builder()
             .with_bind_address(self.server_addr)
-            .with_identity(&self.certificate)
+            .with_identity(self.certificate)
```
Nul-led, line 61: why was this changed?

RJ: wtransport version bump in recent commit changed the api

Nul-led (approved):

cBournhonesque:
Thanks!

Issue #702, opened 2024-11-21
Inputs are not being registered correctly by PrePredicted entities?


cBournhonesque:
In examples client_replication or avian_physics, it looks like inputs are being sent to the remote entity but are not being sent/used correctly for the client that is issuing the inputs.

cBournhonesque:
Client_replication bug fixed in https://github.com/cBournhonesque/lightyear/pull/705

cBournhonesque:
The avian_physics bug happens only on collisions, and it is somehow related to corrections. Disabling corrections completely fixes the issue.

Also there are still rollbacks on collisions between a player and the ball, which is not expected.

PR #713, opened 2024-11-27
Fix pre predicted replication

cBournhonesque:
Subtle bug because the `local_entities` (that are needed to keep the Confirmed tick up-to-date for a given ReplicationGroup) was not being updated for the PrePredicted confirmed entity on the client, because the server's message doesn't count as Spawn (since the entity already exists).

Fixed by manually updating `local_entities` in the case of PrePrediction.
Fixes https://github.com/cBournhonesque/lightyear/issues/702

### DIFF
--- a/examples/avian_physics/src/protocol.rs
+++ b/examples/avian_physics/src/protocol.rs
@@ -43,12 +43,7 @@ impl PlayerBundle {
             id: PlayerId(id),
             position: Position(position),
             color: ColorComponent(color),
-            replicate: client::Replicate {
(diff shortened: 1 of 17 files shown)


examples/avian_physics/src/protocol.rs
```diff
@@ -43,12 +43,7 @@ impl PlayerBundle {
             id: PlayerId(id),
             position: Position(position),
             color: ColorComponent(color),
-            replicate: client::Replicate {
```
cBournhonesque, line 46: Not needed since this is client->server. The replication_group is needed for server->client

PR #716, opened 2024-11-30
Refactor of  avian_3d_physics

Sirmadeira:
So this is the current pull request that is trying to fix the avian_3d_physics sample. 

Current issues are:

1. Rollbacks with single client when giving an input - Half solved - I just increased input delay and that did the tricks but it seens that input buffer from leafwing is broken.Probable cause, Input delay if small, grabs ticks before start tick.
2. Simple collisions with other predicted entities in same client causes rollbacks continuously - Dont know if the current amount is expected or we should worry about that.
3. Add cleaner system - Remove entities from  disconnected client -TODO
4. Add extras forces - Check out other forces like restitution, friction, gravity check if everything work as expected -TODO
5. Make physics debugger work with avian

### DIFF
--- a/examples/simple_box/src/server.rs
+++ b/examples/simple_box/src/server.rs
@@ -65,7 +65,7 @@ pub(crate) fn handle_connections(
 ) {
     for connection in connections.read() {
         let client_id = connection.client_id;
-        // server and client are running in the same app, no need to replicate to the local client
+        // in host-server mode, server and client are running in the same app, no need to replicate to the local client
         let replicate = Replicate {
             sync: SyncTarget {
                 prediction: NetworkTarget::Single(client_id),
--- a/lightyear/src/client/networking.rs
+++ b/lightyear/src/client/networking.rs
@@ -19,6 +19,7 @@ use crate::client::run_conditions::is_disconnected;
 use crate::client::sync::SyncSet;
 use crate::connection::client::{ClientConnection, ConnectionState, DisconnectReason, NetClient};
 use crate::connection::server::IoConfig;
+use crate::prelude::server::ServerConnections;
 use crate::prelude::{
     is_host_server, ChannelRegistry, MainSet, MessageRegistry, TickManager, TimeManager,
 };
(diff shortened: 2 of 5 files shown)


Sirmadeira:
@cBournhonesque  So the problem seens to lie in input delay, when we add a too low input delay. It seens we keep receiveing a  tick lower than the start tick in the buffer which keeps causing rollbacks because well we actually never set a new tick. So if a user exagerates in input delay, rollback forever basically.

Sirmadeira:
It also seens like gravity seriourly cause excessive rollbacks perhaps it is a persistent force?

cBournhonesque:
> @cBournhonesque So the problem seens to lie in input delay, when we add a too low input delay. It seens we keep receiveing a tick lower than the start tick in the buffer which keeps causing rollbacks because well we actually never set a new tick. So if a user exagerates in input delay, rollback forever basically. You can check this out in my current commit just diminish the input delay by 6 in avian_3d_physics and notice how everything breaks.
> 
> BTW I am not sure completely I am not very well versed in your code base, but I would love a slight intro to it. I dont mind help mantaining

Do you have more information about this? It would be helpful so that I can also diagnose the problem.
Are you saying that the server's ActionState corresponds to a tick that is earlier than start tick in the InputBuffer?

Sirmadeira:
Okay here it goes full explanation two current serious bugs affect avian_3d_physics.

First - Client rollbacks with one sole client connected - This was one solved, by removing gravity and sleeping. Not ideal solution but I oughta to do the whole rollback resource and sync components when it comes to those two. If it persists the issue will certainly lie with warm collisions.

Second - Excessive amount of rollbacks when second client connects or moves - The issue here lies in input buffer  specifically in this function 

```
    pub(crate) fn set(&mut self, tick: Tick, value: &ActionState<T>) {
        let Some(start_tick) = self.start_tick else {
            // initialize the buffer
            self.start_tick = Some(tick);
            self.buffer.push_back(BufferItem::Data(value.clone()));
            return;
        };

        // cannot set lower values than start_tick
        if tick < start_tick {
            warn!("Your given tick {:?} is lower than start tick {:?}",tick,start_tick);
            return;
        }

        let end_tick = start_tick + (self.buffer.len() as i16 - 1);
        if tick > end_tick {
            // TODO: Think about how to fill the buffer between ticks
            //  - we want: if an input is missing, we consider that the user did the same action (RocketLeague or Overwatch GDC)

            // TODO: think about whether this is correct or not, it is correct if we always call set()
            //  with monotonically increasing ticks, which I think is the case
            //  maybe that's not correct because the timing information should be different? (i.e. I should tick the action-states myself and set them)
            // fill the ticks between end_tick and tick with a copy of the current ActionState
            for _ in 0..(tick - end_tick - 1) {
                trace!("fill ticks");
                self.buffer.push_back(BufferItem::SameAsPrecedent);
            }
            // add a new value to the buffer, which we will override below
            self.buffer.push_back(BufferItem::Absent);
        }

        // check if the value is the same as the precedent tick, in which case we compress it
        let mut same_as_precedent = false;
        if let Some(action_state) = self.get(tick - 1) {
            if action_state == value {
                same_as_precedent = true;
            }
        }

        // safety: we are guaranteed that the tick is in the buffer
        let entry = self.buffer.get_mut((tick - start_tick) as usize).unwrap();

        if same_as_precedent {
            *entry = BufferItem::SameAsPrecedent;
        } else {
            *entry = BufferItem::Data(value.clone());
        }
    }

```
As you can see we only set new ActionState value, in bufffer. IF tick > start tick.  Problem is, with a low minimum_input_delay_ticks, like 0 for example server keeps being passed a tick lower than start tick. Almost continously, which means he doesnt set new data value he persists in precedent. Which causes teleportation in clients because well server doesnt move his entity.
Sample video .
[Gravação de tela de 2024-12-01 17-46-55.webm](https://github.com/user-attachments/assets/46fec4a1-f187-4af8-9a12-de866fb79358)

PR #732, opened 2024-12-17
Cargo refactor & examples work

RJ:
Refactor Cargo workspace and examples

(Precursor work to hosting live examples on edgegap)

* moved all dependencies to top-level Cargo file, under workspace.dependencies. this helps ensure we are always using the same version of deps that are needed across multiple crates (eg lightyear, tests, examples/*).
* changed bevy dep to, by default, only add features needed for headless operation
* added "gui" feature to lightyear_examples_common and examples, which enables the other gui-related bevy features. this allows us to build true headless server binaries for the examples.
* refactor some setup code in common, so examples explicity say if the plugin is for client/server/shared/renderer. This is so we can omit UI related stuff in headless builds.
* added a rendering plugin in common that sets up the window plugin, a connect button system, and some status text, which all the examples will use. individual examples still responsible for adding a camera bundle and rendering their game.
* added bevygap features to common and simple_box and spaceships examples. the example games don't need to write any bevygap code, they just need to enable the appropriate feature on lightyear_examples_common and the behaviour of the connect system in the common renderer plugin changes accordingly.
* added a common/BevygapSharedExtensionPlugin plugin that replicates a server metadata resource and displays the FQDN and location of the server you connect to.
* added some feature combo guards in common/lib. so now you can be sure that (eg) if the steam feature is enabled, you can't be on wasm. this simplifies some of the cfg checks in the code (of which there are now more..).
* added build system for bevygap-enabled wasm frontend for examples, publishing to github pages.
* added build system for bevygap-enabled single docker image for all server binaries, publishing to gh container registry.
* fixed a few minor linting things as i noticed them
* wrote and used arbctl to manipulate docker version tags on deployed edgegap applications as part of build pipeline
* made and migrated to lightyear org on my edgegap account.
* fixed benches and some tests
* disabled broken native websocket test
* Note that `metrics` feature doesn't compile (unchanged from main)

## TODO in follow-up PR

* add egui world inspector to common renderer with shortcut to toggle
* fix example ui that uses diagnostics overlay from overlapping with common ui. connect button kinda hidden by metrics.
* update the other examples to support bevygap and split client/server features
* either write a template for the example menu index.html, or maybe add to the existing book via iframes pointing to ghpages?
* decide where to run the matchmaker bits, at what url. currently sharing infra with my bevygap-spaceships demo.
* ideally the wasm and server builders should be two phase - compile artifacts, once both are known to compile, then deploy.. want to avoid situation where we push new servers and still have old wasm clients due to failed build, and visa-versa. the "right" way might be to embed the application version edgegap uses, but that adds too much complexity for now, for some simple examples.


RJ:
After merging please disable the 2 new workflows for server and wasm builds.

PR #734, opened 2024-12-20
add missing certificate bit to examples' settings.rons

RJ:
restore original examples behaviour of using the certificate pem files, by adding the required bit to settings.ron for examples


cBournhonesque:
Thanks for the effort; I had already changed it on my migration branch though!

PR #753, opened 2025-01-01
Allow public access to entity mapping function

BigBadE:
When sending messages to the server, users have no way to map local to remote entities and vice versa. This means there's no built-in way to make messages store entity references. This could be solved by adding a component to the entity instead, but it's just such a more complex and convoluted solution to a simple issue. There also doesn't seem to be a reason why these constant non-state changing functions in a publicly accessible struct aren't able to be read.

### DIFF
--- a/lightyear/src/shared/replication/entity_map.rs
+++ b/lightyear/src/shared/replication/entity_map.rs
@@ -88,7 +88,7 @@ impl RemoteEntityMap {
     /// It's possible that the remote_entity was already mapped by the sender,
     /// in which case we don't want to map it again
     #[inline]
-    pub(crate) fn get_local(&self, remote_entity: Entity) -> Option<Entity> {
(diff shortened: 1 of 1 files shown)


cBournhonesque:
These functions are heavily tied to lightyear internals and should probably stay private.

Entities within messages are actually automatically mapped to the correct World when a message is sent/received, so there shouldn't be a need to expose these functions. Did you run `add_map_entities()` when registering the message to your protocol? https://docs.rs/lightyear/latest/lightyear/prelude/struct.MessageRegistry.html#customizing-message-behaviour

Or is there any situation in particular where the automatic entity mapping doesn't work for you?

BigBadE:
Oh wow, I missed that feature! I was wondering how there wasn't some way to map entities that I could find, I guess I just searched in the wrong place.

PR #776, opened 2025-01-06
fix: register prespawned entity in predicted_entity_map during server…

OlivierCoue:
Currently when a server pre spawned entity is matched with a client one, PreSpawnedPlayerObject is removed and Predicted is added but it does not get added to the predicted_entity_map.

This is an issue because the Predicted entity is not despawned when the confirmed one is despawned (by despawn_confirmed) 

https://github.com/cBournhonesque/lightyear/blob/774e948dcadc2564ea07540b5ea73ed8cbd81368/lightyear/src/client/prediction/despawn.rs#L90


cBournhonesque:
Thanks, great catch!

Issue #683, opened 2024-11-02
game controller crash on WASM


simbleau:
I have a standard xbox controller and I'm on MacOS. My game runs on wasm.

I can reproduce a panic in lightyear's integration with leafwing input manager:
<img width="1467" alt="image" src="https://github.com/user-attachments/assets/4bc4b4da-51c2-4f6f-bf7f-604993a20c49">

Unfortunately I don't know much other than this, console is pretty unhelpful.

Steps to reproduce:
- Run a bevy game
- Connect a new xbox controller after the game starts
- Panic occurs

If I have the xbox controller connected prior to the page (bevy game) loading, I don't get a panic.

cBournhonesque:
I;m nit sure, it might be a leafwing issue?

Jieiku:
One way to test would be to see if this happens in a bevy game that uses leafwing but NOT lightyear.

If I get a little free time I will try to reproduce this and determine the cause.

Jieiku:
@simbleau I added basic controller support to the avian 3d character example: https://github.com/cBournhonesque/lightyear/commit/874c54af56a9bcfd7af7c89de68cfcbe2dab698c

I am on linux.

- controller unplugged
- launched the client:

```bash
cd lightyear/examples/avian_3d_character
trunk serve --features=client
```

- launch the server:

```bash
cd lightyear/examples/avian_3d_character
cargo run --features=server
```

- open google chrome to the address that trunk serve shows: `http://127.0.0.1:8080/`

- character seems to move fine with WASD keys.
- plug controller in, character also works fine with controller.

I am guessing your issue was either resolved in leafwing since you opened this a couple months ago, or your issue is specific to your operating system or browser.... I would think it should not really matter what your OS is if you are also using google chrome.

cBournhonesque:
I'll close this for now; will re-open if we can get another reproduction

PR #792, opened 2025-01-09
add missing dependencies, add avian3d simple gamepad support

Jieiku:
This pull request adds a few missing dependencies.

benches/bitcode_packing was missing depencies, which was flagging some errors in my IDE.

The build_wasm.sh script in the examples folder was failing because it could not find the websocket dependency, so I added that dependency to the simple_box, spaceships, and avian_3d_character examples, so that I could build the wasm.

I also added very basic gamepad support to avian_3d_character, which provides example gamepad usage and will allow testing for https://github.com/cBournhonesque/lightyear/issues/683

### DIFF
--- a/examples/spaceships/Cargo.toml
+++ b/examples/spaceships/Cargo.toml
@@ -20,7 +20,7 @@ bevygap_server = ["lightyear_examples_common/bevygap_server", "server"]
 steam = ["lightyear_examples_common/steam"]
 
 [dependencies]
-lightyear_examples_common = { workspace = true, features = ["websocket"] }
+lightyear_examples_common.workspace = true
 metrics-exporter-prometheus = { workspace = true, optional = true }
 leafwing-input-manager = { workspace = true, features = ["keyboard", "mouse"] }
 avian2d = { workspace = true, features = [
--- a/examples/spaceships/Cargo.toml
+++ b/examples/spaceships/Cargo.toml
@@ -20,7 +20,7 @@ bevygap_server = ["lightyear_examples_common/bevygap_server", "server"]
 steam = ["lightyear_examples_common/steam"]
 
 [dependencies]
-lightyear_examples_common.workspace = true
+lightyear_examples_common = { workspace = true, features = ["websocket"] }
 metrics-exporter-prometheus = { workspace = true, optional = true }
 leafwing-input-manager = { workspace = true, features = ["keyboard", "mouse"] }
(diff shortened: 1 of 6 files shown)


examples/spaceships/Cargo.toml
```diff
@@ -20,7 +20,7 @@ bevygap_server = ["lightyear_examples_common/bevygap_server", "server"]
 steam = ["lightyear_examples_common/steam"]
 
 [dependencies]
-lightyear_examples_common.workspace = true
+lightyear_examples_common = { workspace = true, features = ["websocket"] }
```
cBournhonesque, line 23: that shouldn't be necessary since the `default` features above contain websocket, no?
```
default = ["lightyear_examples_common/websocket"]
```

Jieiku: not necessary to run the example with cargo, but when you run the build_wasm script shell script in the examples folder it specifically says that it cannot find the WebSocket Variant in ServerTransports, once I added that feature to common the build_wasm script was able to compile the wasm examples.

Jieiku: I am building again with it commented out so that I can screen shot the error.

cBournhonesque: I see, i'm a little hesitant on including this dependency by default for the example.
Maybe the build_wasm script command can just use 
`--features=lightyear_examples_common/websocket`

(I haven't used the script myself, it was contributed by @RJ for bevygap, I just call `trunk server` to start the wasm examples)

Jieiku: ahhh, that would work too! I will add another commit in a second.

Jieiku:
Tell me more about the `trunk server` that you use to test the wasm examples. I would like to try that out.

examples/spaceships/Cargo.toml
Jieiku: The error about the websockets was from running [./build_wasm.sh](https://github.com/cBournhonesque/lightyear/blob/main/examples/build_wasm.sh)



I will try out using `trunk server` like you do:

https://github.com/cBournhonesque/lightyear/tree/main/examples#testing-in-wasm-with-webtransport

Jieiku: @cBournhonesque let me know if anything else needs fixed/changed. Thanks :)

cBournhonesque:
LGTM, thanks!

PR #841, opened 2025-01-26
Separate new() and new_with_app_id()

NathanTaylorHunt:
If you call init() on the SteamAPI, it will automatically determine the AppId, otherwise you can call init_app() and provide an explicit AppId.

https://docs.rs/steamworks/latest/steamworks/struct.Client.html#method.init

### DIFF
--- a/lightyear/src/connection/steam/steamworks_client.rs
+++ b/lightyear/src/connection/steam/steamworks_client.rs
@@ -19,13 +19,25 @@ impl std::fmt::Debug for SteamworksClient {
 }
 
 impl SteamworksClient {
-    /// Creates and initializes the Steamworks client. This must only be called
-    /// once per application run.
-    pub fn new(app_id: u32) -> Self {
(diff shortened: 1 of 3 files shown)


lightyear/src/connection/steam/steamworks_client.rs
```diff
@@ -31,6 +31,18 @@ impl SteamworksClient {
         }
     }
 
+    /// Creates and initializes the Steamworks client. This must only be called
+    /// once per application run.
```
cBournhonesque, line 35: Can you add a comment that the app_id will usually be read from a steam_appid.txt file?

NathanTaylorHunt: Sure, I copied the comment from the steamworks rs docs:
https://docs.rs/steamworks/latest/steamworks/struct.Client.html#method.init

cBournhonesque:
Thanks!

PR #844, opened 2025-01-26
Emit disconnect event on connection failure

cBournhonesque:
We were emitting a DisconnectEvent on the client if we were already connected and a disconnect happened (from io, or from the server, or because the client request it).

However we were not emitting a disconnect-event when the client fails to connect (for example for io reasons, etc.)


lightyear/src/client/networking.rs
```diff
@@ -66,14 +66,6 @@ impl Plugin for ClientNetworkingPlugin {
                     .chain(),
             )
             // SYSTEMS
-            .add_systems(
```
cBournhonesque, line 69: this system was weirdly duplicated + i don't think it's necessary anymore for it to be in a different set

PR #845, opened 2025-01-26
fixup disconnect events firing, adapt steam disconnections to ConnectionError

Piefayth:
Looks like the steamworks `NetConnectionEnd` type was getting mapped to the now removed `DisconnectReason` type. So I just gave it an entry in `ConnectionError`. I can do that differently if you had something else in mind


cBournhonesque (approved):
Thanks!

PR #854, opened 2025-01-28
Improve error handling and reporting for Netcode

Piefayth:
- Adds several entries to the `netcode::Error` type to replace some log statements / throwaway failures
- Many errors in netcode now bubble where it makes sense for them to
- `NetcodeServer` now tracks a `Vec<ConnectionError>` of `client_errors`. This uses the less specific `ConnectionError` instead of the more specific `netcode::Error` so we can treat them generically later. It might make more sense to put this directly on `ServerConnection`, but it looked like it might've been fiddly work to replace the enum with the struct. What do you think?


lightyear/src/server/networking.rs
```diff
@@ -295,15 +312,59 @@ pub(crate) fn send(
...
+        }
+    }
 }
 
+fn react_to_client_error(
+    connection_manager: &mut ResMut<ConnectionManager>,
+    server: &mut ServerConnection,
+    error: ConnectionError
+ ) {
+    let fatal_error_client_info = match &error {
```
Piefayth, line 339: Thumbing through the error tree to dig out fatal errors sucks.

lightyear/src/server/networking.rs
```diff
@@ -295,13 +320,53 @@ pub(crate) fn send(
...
+                _ => None,
+            },
+            _ => None,
+        },
+        _ => None,
+    };
+    if let Some((client_id, err)) = fatal_error_client_info {
+        error!("Fatal Client Error: {:?}", error);
+        let _ = server.disconnect(client_id);
+        connection_manager.remove(client_id);
```
Piefayth, line 366: Is there a better way to disconnect? `ServerCommands` is bad because then callers to `send` have to provide a `Commands`

lightyear/src/connection/netcode/server.rs
```diff
@@ -970,7 +985,7 @@ impl<Ctx> NetcodeServer<Ctx> {
     /// The server will send a number of redundant disconnect packets to the client, and then remove its connection info.
     pub(crate) fn disconnect_by_addr(&mut self, addr: SocketAddr, io: &mut Io) -> Result<()> {
         let Some(client_id) = self.conn_cache.client_id_map.get(&addr) else {
-            return Err(Error::ClientNotFound);
+            return Err(Error::ClientNotConnected);
```
cBournhonesque, line 988: Should ClientNotConnected contain either id or addr info?

lightyear/src/connection/netcode/server.rs
```diff
@@ -898,7 +913,7 @@ impl<Ctx> NetcodeServer<Ctx> {
     pub fn send_all(&mut self, buf: &[u8], io: &mut Io) -> Result<()> {
         for id in self.conn_cache.ids() {
             match self.send(buf, id, io) {
-                Ok(_) | Err(Error::ClientNotConnected) | Err(Error::ClientNotFound) => continue,
+                Ok(_) | Err(Error::ClientNotConnected) | Err(Error::ClientNotFound(_)) => continue,
```
cBournhonesque, line 916: Maybe i missed it, but in case of an error don't we want to call `self.handle_client_error`?

This method doesn't seem used at all, anyway

lightyear/src/connection/netcode/server.rs
```diff
@@ -1172,6 +1189,7 @@ pub(crate) mod connection {
                 server,
                 io_config,
                 io: None,
+                client_errors: vec![],
```
cBournhonesque, line 1192: This doesn't seem used

lightyear/src/connection/netcode/server.rs
```diff
@@ -1037,6 +1052,7 @@ pub(crate) mod connection {
     #[derive(Resource)]
     pub struct Server {
         pub(crate) server: NetcodeServer<NetcodeServerContext>,
+        client_errors: Vec<ConnectionError>,
```
cBournhonesque, line 1055: this doesn't seem used

lightyear/src/connection/netcode/server.rs
```diff
@@ -384,6 +385,7 @@ pub struct NetcodeServer<Ctx = ()> {
     conn_cache: ConnectionCache,
     token_entries: TokenEntries,
     cfg: ServerConfig<Ctx>,
+    pub client_errors: Vec<ConnectionError>,
```
cBournhonesque, line 388: Do we need to make this pub?

lightyear/src/server/networking.rs
```diff
@@ -103,6 +106,8 @@ pub(crate) fn receive_packets(
     time_manager.update(delta);
     trace!(time = ?time_manager.current_time(), tick = ?tick_manager.tick(), "receive");
 
+    let mut client_errors = Vec::new();
```
cBournhonesque, line 109: Can we use `Local` system param to avoid allocating a new vec every frame?

lightyear/src/server/networking.rs
```diff
@@ -170,6 +189,12 @@ pub(crate) fn receive_packets(
         }
     }
 
+    for (server_idx, error) in client_errors {
```
cBournhonesque, line 192: We do this here and not above because of borrow checker issues?

lightyear/src/server/networking.rs
```diff
@@ -139,13 +144,27 @@ pub(crate) fn receive_packets(
...
+                            .drain(..)
+                            .map(|e| (server_idx, e)),
+                    );
+                }
+                #[cfg(all(feature = "steam", not(target_family = "wasm")))]
+                ServerConnection::Steam(_) => {} // todo: steam errors
+            };
         }
+
+        let new_disconnections = netserver.new_disconnections();
```
cBournhonesque, line 167: Did you move this here because of borrow-checker issues?

lightyear/src/connection/netcode/server.rs
cBournhonesque: Ah it's because for `send` there is no buffering so we can just return the error directly?

lightyear/src/server/networking.rs
cBournhonesque: Yes, ideally we could create a trait `Recoverable` and make the various errors 
(IoError, NetcodeError, TransportError) implement it

lightyear/src/server/networking.rs
cBournhonesque: I'd prefer to just add `Commands` to `send` so that the disconnection logic is centralized in a single place. But yes maybe I should have `disconnect_internal` that just requires `ConnectionManager` and `ServerConnections`

lightyear/src/transport/udp.rs
```diff
@@ -98,11 +98,19 @@ pub struct UdpSocketBuffer {
...
-            .as_ref()
-            .lock()
-            .unwrap()
-            .send_to(payload, address)?;
+        let socket = self.socket.as_ref().lock().unwrap();
+
+        #[cfg(target_os = "windows")]
+        {
+            let mut peek_buf = [0u8; 1];
+            match socket.peek_from(&mut peek_buf) {
```
cBournhonesque, line 106: That's great, could you please add a comment explaining your findings (e.g. that on windows the error happens in send and not receive, etc.) and what this code is suppsoed to fix? (the logspam if the remote udp receiver closes the socket)

lightyear/src/connection/netcode/server.rs
Piefayth: that's right

lightyear/src/connection/netcode/server.rs
Piefayth: vestigial, yep 👍

lightyear/src/connection/netcode/server.rs
Piefayth: it's accessed outside of `netcode`; do you want an accessor method instead?

lightyear/src/server/networking.rs
Piefayth: Yeah, borrow checker issues abound trying to collect the things I need to disconnect. Maybe it's better with ServerCommands for disconnects, I'll see if I can revert some of the rearrangement when I do that

lightyear/src/server/networking.rs
Piefayth: i'm gonna punt on that until we've identified more than one fatal error, if that's okay

lightyear/src/connection/netcode/server.rs
cBournhonesque: It looks like the client_errors is only used for the `receive` side of things.
Maybe the `try_update` function of the `NetServer` trait could be updated to something like:
```
fn try_update(&mut self, delta_ms: f64) -> Result<Vec<ConnectionError>, ConnectionError>;
```

(the first connection-errors are those that come from connections, the second one is from errors in `try_update` itself.
Or maybe just
```
fn try_update(&mut self, delta_ms: f64) -> Vec<ConnectionError>;
```
if it's too confusing?

Then there's a bit less dancing around and we can just handle the errors directly after update time

cBournhonesque:
Thank you, this is great!

PR #862, opened 2025-02-01
fix ReplicateOnce components not being replicated to new clients

Piefayth:
just don't return early from this function so the new connection logic actually runs


cBournhonesque:
Looks great, thanks!
added a unit test

PR #914, opened 2025-02-19
Getting Crashes from these lines

SueHeir:
Feel free to ignore this if these crashes are the fault of the user (Me), just ran into them and thought you should take a look. The leafwingUserAction fixes probably aren't helpful because you already redid that. The handle_pre_predicted error is coming from this bare-bone menu [example,](https://github.com/SueHeir/lightyear-menu/issues/2) and seems to happen randomly, more so when disconnecting and reconnecting.


cBournhonesque:
Thanks! Instead of merging now I will try to understand why these lines are reachable, because I was under the impression that they were not

cBournhonesque:
Does this not happen anymore?

PR #945, opened 2025-03-09
Fixes for bevy-main wrt latest bevy

alexgershberg:
I was trying to consume bevy-main branch in my own project, so I figured it's useful to share the fixes I've made to that branch.

that said, i couldn't quite figure out what's going on with PreSpawnedPlayerObject, so the tests that use it fail to compile (was this struct removed?). Apart from that, there should be no more compilation errors.

I also ran cargo fmt and fixed a few deprecated functions (like: send_batch -> write_batch, send -> write)

### DIFF
--- a/lightyear/src/client/input/mod.rs
+++ b/lightyear/src/client/input/mod.rs
@@ -235,22 +235,22 @@ fn get_non_rollback_action_state<A: UserActionState>(
     mut action_state_query: Query<(Entity, &mut A, &InputBuffer<A>)>,
 ) {
     let tick = tick_manager.tick();
-    // for (entity, mut action_state, input_buffer) in action_state_query.iter_mut() {
-    //     // We only apply the ActionState from the buffer if we have one.
-    //     // If we don't (which could happen for remote inputs), we won't do anything.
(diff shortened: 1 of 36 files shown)


lightyear/src/client/input/mod.rs
```diff
@@ -235,22 +235,22 @@ fn get_non_rollback_action_state<A: UserActionState>(
...
-                ?entity,
-                ?tick,
-                "fetched action state {:?} from input buffer: {:?}",
-                action_state,
-                // action_state.get_pressed(),
-                input_buffer
-            );
-        }
-    }
+    // for (entity, mut action_state, input_buffer) in action_state_query.iter_mut() {
```
alexgershberg, line 238: Completely forgot about this...

lightyear/src/client/prediction/rollback.rs
```diff
@@ -1548,7 +1550,7 @@ mod unit_tests {
         let _ = stepper
             .client_app
             .world_mut()
-            .run_system_once(check_rollback::<ComponentSyncModeFull>);
+            .run_system_once(check_rollback);
```
cBournhonesque, line 1553: this test might not work, i'm commenting to remember

PR #951, opened 2025-03-12
Make SteamworksClient construction fallible

OleStrohm:
When Steam is not running the creation of the steamworks::Client will fail, and as-is that causes an unwrap in the constructor of SteamworksClient. The way to work around it would be to create the SteamworksClient manually, but this is not possible since the fields are private. Therefore I'm suggesting to make the construction fallible so you can choose to construct it yourself, and just not turn on the steam server if the construction fails.

I've left the existing behaviour of the built-in fallback SteamworksClient unwrapping on initialization error, but maybe this should be addressed too.

### DIFF
--- a/lightyear/src/connection/client.rs
+++ b/lightyear/src/connection/client.rs
@@ -156,9 +156,9 @@ impl NetConfig {
             } => {
                 let client = super::steam::client::Client::new(
                     steamworks_client.unwrap_or_else(|| {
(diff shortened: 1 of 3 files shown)


cBournhonesque:
> I've left the existing behaviour of the built-in fallback SteamworksClient unwrapping on initialization error, but maybe this should be addressed too.

Yep definitely there should be error-handling there as well, but this works for now. Thanks!

PR #953, opened 2025-03-13
Visualizer config bool for separate modes

SueHeir:
Hello again!

I ran into an issue where I my "headless" server thread app runs a "window" (without displaying anything) to keep a lot of different plugins functioning (many plugins don't consider headless).  This flag allows you disable the visualizer for the server thread while having the visualizer feature work as aspected for the client main thread.  Without this the visualizer crashes the server thread.


lightyear/src/client/networking.rs
```diff
@@ -266,7 +266,7 @@ pub(crate) fn sync_update(
         {
             metrics::counter!("sync::resync_event").increment(1);
         }
-        debug!("Triggering TickSync event: {tick_event:?}");
```
cBournhonesque, line 269: These need to be reverted

cBournhonesque:
I'm not sure I fully understand the issue. 
If your server is displaying a window, then the visualizer shouldn't crash right? I think the visualizer just crashes if you try to run it in headless mode.
Maybe a simpler fix could be to check if some `WindowPlugin` or `RenderPlugin` is running. If not, do not use the visualizer. What do you think?

PR #991, opened 2025-04-18
<Book> Add hidden lines to setup.md as well as a heads-up

SadraMoh:
- Add hidden lines to the `setup.md` example that show the implementation of `Add` and `Mul` for `PlayerPosition`. Trying the example without the hidden lines causes compile errors.
- Also add a small hint at the bottom of the example.  

The whitespace changes are caused by my formatter.


cBournhonesque:
Thank you!

PR #994, opened 2025-04-26
cb/0.20 - fix button clicks, and tweaks for common client renderer

RJ:
doesn't fix the button problem, but i changed the spawning to use the new 0.16 style, and you can now press F12 and expand the entity tree to see that yes, the Interaction component for the button entity does change between Hovered and Clicked as expected. Still not sure why the Pointer<Click> event doesn't fire with a target.


RJ:
update: fixed clicks.. we need to add the ui picking backend via cargo feature.

PR #1078, opened 2025-07-06
skip crossbeam links in ServerUdpIO

SueHeir:
I believe this bug will show up for steamIO and any other future IO's

The send vec was being drained by Udp for a crossbeam link

### DIFF
--- a/lightyear_crossbeam/src/lib.rs
+++ b/lightyear_crossbeam/src/lib.rs
@@ -39,8 +39,8 @@ const LOCALHOST: SocketAddr = SocketAddr::new(core::net::IpAddr::V4(Ipv4Addr::LO
 /// and receiver ends of the channels.
 #[derive(Component, Clone)]
 #[require(Link::new(None))]
-#[require(LocalAddr(LOCALHOST))]
-#[require(PeerAddr(LOCALHOST))]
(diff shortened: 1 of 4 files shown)


SueHeir:
Because the server component stores every link entity we can't give an error message when Udp finds a link that is not a Udp link.  Maybe you'll want to restructure more stuff later:) but I figured this would be good for now!

SueHeir:
This breaks Udp connections

SueHeir:
Ok its working now

lightyear_crossbeam/src/lib.rs
```diff
@@ -39,8 +39,8 @@ const LOCALHOST: SocketAddr = SocketAddr::new(core::net::IpAddr::V4(Ipv4Addr::LO
 /// and receiver ends of the channels.
 #[derive(Component, Clone)]
 #[require(Link::new(None))]
-#[require(LocalAddr(LOCALHOST))]
-#[require(PeerAddr(LOCALHOST))]
+// #[require(LocalAddr(LOCALHOST))]
```
cBournhonesque, line 42: I think we might still want those, no?
If the user provides their own LocalAddr and PeerAddr, then these won't be inserted

PR #1085, opened 2025-07-10
Fix RemoteTimeline syncing

Avilad:
Two main issues:
1. InputTimeline drives `Time<Virtual>` and speeds up or slows down to attempt to synchronize with RemoteTimeline. Advancing RemoteTimeline based on `Res<Time>` (aka `Res<Time<Virtual>>`) means when we're, say, a bit behind the server and speed up the input timeline to catch up, we also start updating our estimate of the server timeline as if the server was speeding up along with us. I noticed that with more extreme SyncConfig settings, this was sometimes causing runaway feedback loops, where a client kept accelerating and getting further and further ahead of the server, while thinking the server was still ahead of it. Even with more conservative settings and no runaway feedback loop, it's probably not helping the synchronization.
TL;DR yes imo this should be based on real time

2. I chased my tail for a while thinking that how we track `this.offset` in RemoteTimeline is somehow wrong, because it seemed like the tracking worked better when I updated `this.now` directly with corrections instead of maintaining a separate `now` and `offset`. But it turns out that `TickDelta::add` was just broken for negative RHS 😛

I do think `RemoteTimeline::update` could be simplified (while still behaving exactly the same) by getting rid of the `offset` and just adjusting `now` based on `raw_offset`, but I don't know if you had a specific reason to keep those separate that I'm not aware of. I also wonder if TickDelta math could be simplified (and made less branchy) if `tick_diff` was a signed integer instead of having a separate `neg: bool` member. What do you think?

(also fixed typo in hierarchy.rs that broke compilation)

### DIFF
--- a/lightyear_core/src/time.rs
+++ b/lightyear_core/src/time.rs
@@ -395,6 +395,9 @@ impl Add for TickDelta {
     type Output = TickDelta;
 
     fn add(self, rhs: Self) -> Self::Output {
+        if rhs.is_negative() {
+            return self - (-rhs);
+        }
         if self.is_negative() {
             return rhs - (-self);
         }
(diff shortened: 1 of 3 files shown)


cBournhonesque (approved):
Thanks, the fixes are awesome!

PR #1105, opened 2025-07-16
Support immutable components in prediction/interpolation

cBournhonesque:
- This also unblocks deprecating RelationshipSync, since users can now directly do
`app.register_component::<R>` on their relationship components


examples/lobby/src/client.rs
```diff
@@ -49,7 +49,7 @@ impl Plugin for ExampleClientPlugin {
                 .run_if(in_state(AppState::Game)),
         );
         app.add_systems(
-            EguiContextPass,
+            PostUpdate,
```
vladbat00, line 52: `EguiContextPass` got renamed to `EguiPrimaryContextPass`

PR #1121, opened 2025-07-23
Fix `InputPlugin` of `lightyear_inputs_bei` to `FixedPreUpdate`

hukasu:
Having a `InputContext` on `PreUpdate` causes weirdness, to prevent this `InputPlugin` (of `lightyear_inputs_bei`) now only allows `InputContext` that run in `FixedPreUpdate`

### DIFF
--- a/lightyear_inputs_bei/src/plugin.rs
+++ b/lightyear_inputs_bei/src/plugin.rs
@@ -27,7 +27,7 @@ impl<C> Default for InputPlugin<C> {
     }
 }
 
-impl<C: InputContext> Plugin for InputPlugin<C> {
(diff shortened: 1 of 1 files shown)


cBournhonesque:
> Should i replicate this to the other 2 crates?

The other crates don't specify their schedules.  

Do you think adding an unimplemented diagnostic would help? https://doc.rust-lang.org/beta/reference/attributes/diagnostics.html#the-diagnosticon_unimplemented-attribute 

So that users get a helpful message if they forget to specify the correct schedule

hukasu:
Yes, that would be nice

hukasu:
https://doc.rust-lang.org/beta/reference/attributes/diagnostics.html#the-diagnosticon_unimplemented-attribute can only be used on trait definitions

PR #1136, opened 2025-07-29
fix(message-registry): do not add to message registry hasher for duplicate values

komodo472:


cBournhonesque:
Thanks; I think i will go with https://github.com/cBournhonesque/lightyear/pull/1138 instead, and simply not add the `ChecksumSend` plugin if the server feature is enabled (which is the case for host-client mode).

I stole your fix for the movement system, so I added you as co-author!

PR #1140, opened 2025-08-01
Remove noisy logs

magic-sysrq:


cBournhonesque:
Thanks, i keep forgetting to change logs back after enabling them for debugging..

Issue #1142, opened 2025-08-02
`lightyear_replication` build fails


esuriru:
Attempted to build `simple_box`  with rust 1.87; on windows 11 platform, results in 2 errors

Log output:
```
   Compiling lightyear_connection v0.22.5 (D:\lightyear\lightyear_connection)                                                                   
   Compiling lightyear_transport v0.22.5 (D:\lightyear\lightyear_transport)                                                                     
   Compiling xwt-wtransport v0.13.2                                                                                                             
   Compiling bevy_gizmos_macros v0.16.1                                                                                                         
   Compiling bevy_input_focus v0.16.1                                                                                                           
   Compiling bevy_image v0.16.1                                                                                                                 
   Compiling lightyear_messages v0.22.5 (D:\lightyear\lightyear_messages)
   Compiling bevy_winit v0.16.1                                                                                                                 
   Compiling aeronet_webtransport v0.14.0                                                                                                       
   Compiling lightyear_aeronet v0.22.5 (D:\lightyear\lightyear_aeronet)
   Compiling bevy_mesh v0.16.1                                                                                                                  
   Compiling lightyear_sync v0.22.5 (D:\lightyear\lightyear_sync)                                                                               
   Compiling lightyear_replication v0.22.5 (D:\lightyear\lightyear_replication)                                                                 
   Compiling egui v0.31.1                                                                                                                       
error[E0658]: `let` expressions in this position are unstable                                                                                   
   --> lightyear_replication\src\send\buffer_bis.rs:340:33
    |
340 |         if delta_compression && let Some((delta_manager, shared_tick)) = delta {
    |                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #53667 <https://github.com/rust-lang/rust/issues/53667> for more information

error[E0658]: `let` expressions in this position are unstable                                                                                   
   --> lightyear_replication\src\send\buffer_bis.rs:533:41
    |
533 |                 if delta_compression && let Some(delta) = delta {
    |                                         ^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #53667 <https://github.com/rust-lang/rust/issues/53667> for more information

   Compiling webbrowser v1.0.5                                                                                                                  
   Compiling lightyear_netcode v0.22.5 (D:\lightyear\lightyear_netcode)                                                                         
[...]

PR #1143, opened 2025-08-02
Update `rust-version` to 1.88

esuriru:
This should close https://github.com/cBournhonesque/lightyear/issues/1142 (see for more info)

Increase the minimum required rust-version to 1.88; which is the version this issue (https://github.com/rust-lang/rust/issues/53667) suggests.


cBournhonesque:
Thanks!

PR #1162, opened 2025-08-16
History Buffer Causes Rollbacks before any data is entered into History Buffer

SueHeir:
I was getting a single rollback every time I spawned a projectile; history buffers that have no data yet will cause a rollback, when I believe they shouldn't do anything until data is received.

I added a is_new bool on the history buffer that gets set to false once any data is added. If is_new is true, we skip all the checks for rollbacks in the check_rollback function

This seems to fix my projectiles moving forward every time I shoot a projectile, not sure why

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ members = [
   "lightyear_udp",
   "lightyear_utils",
   "lightyear_webtransport",
+  "lightyear_websocket",
   # internal
   "benches/*",
   "demos/*",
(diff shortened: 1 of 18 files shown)


cBournhonesque:
I don't think this is correct.
It's normal that any Predicted entity will trigger a Rollback, this is because the entity was received in the Replicated/Confirmed entity, and we need to bring it to the Predicted entity. If you want to avoid rollbacks, you need to do a predicted Pre-Spawn like in https://github.com/cBournhonesque/lightyear/blob/main/examples/fps/src/shared.rs#L172

PR #1168, opened 2025-08-17
DisableDuringRollback

SueHeir:
Disables all entities with the DisableDuringRollback component during rollbacks

### DIFF
--- a/demos/spaceships/Cargo.toml
+++ b/demos/spaceships/Cargo.toml
@@ -9,11 +9,12 @@ edition.workspace = true
 publish = false
 
 [features]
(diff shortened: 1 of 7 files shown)


cBournhonesque:
Thanks, I think this is a promising approach.
How come your interpolated entities are spawning predicted projectiles? 

So the projectiles can be in the same timeline as the client? Would that mean that the projectiles are automatically spawned 'in front' of the shooter, thanks to the initial rollaback?

cBournhonesque:
Thanks, that's helpful; I will add something similar to what you propose

PR #1259, opened 2025-10-15
Feat: New lag compensation API

Sirmadeira:


cBournhonesque:
Can you please explain what are the shortcomings with the current API that this aims to address?

PR #1310, opened 2025-11-17
Feat: Taplo format could add CI if approved

Sirmadeira:


Sirmadeira:
@cBournhonesque taplo format in this case, sure thing

PR #1312, opened 2025-11-17
Feat: Readition of lightyear serde features

Sirmadeira:


lightyear_serde/src/lib.rs
```diff
@@ -36,6 +36,11 @@ pub mod writer;
 
 /// Commonly used items from the `lightyear_serde` crate.
 pub mod prelude {
+    pub use crate::entity_map::*;
```
cBournhonesque: What types among these do you need?
I would rather users import them directly rather than via the prelude, as some types might conflict with std types. For example 'Reader'

PR #1320, opened 2025-11-19
fix: keep InputTimeline in sync with LocalTimeline

Avilad:
The timeline sync algorithm in InputTimeline uses `self.now()` to calculate the error that drives the adjustment of the game clock / LocalTimeline, but LocalTimeline isn't in that control loop - so even though input and local typically stay in sync because they're incremented by the same delta each frame (LocalTimeline's overstep is "stored" in `Time<Fixed>.overstep_fraction()`), if they happen to drift the client doesn't recover from it, and since the InputTimeline is the one that's synchronized to sync_objective(), the client's simulation ends up offset from the intended tick by the amount of drift that has built up.

And this happens in practice! The server stall stress test I added to Spaceships helps to expose the problem. I noticed it quickly when experimenting with fixed-point tick types because of the reduced precision, but I was able to reproduce it on main as well - you can check out #89ff997 and see it with your own eyes. I took some screenshots, first a baseline (shortly after starting the server and both clients):
Screenshot 2025-11-18 at 7 11 03 PM
And after running all 3 for ~30 minutes with SPACESHIPS_SERVER_STALL_MS=2000 and SPACESHIPS_SERVER_STALL_INTERVAL_TICKS=1000, this is what the `num_buffered_inputs` labels were stabilizing at (just before each stall):
Screenshot 2025-11-18 at 7 40 10 PM

 `InputTimeline::advance_timeline` is correct with infinite precision, but it doesn't do the exact same operations as [run_fixed_main_schedule](https://github.com/bevyengine/bevy/blob/d87d491d29a628db44a69c04665c56322d912512/crates/bevy_time/src/fixed.rs#L243): advance_timeline calls apply_duration which converts the delta to an f32 and divides by tick_duration to get a TickDelta, then adds that to itself; run_fixed_main_schedule instead adds the delta as a Duration to an internal accumulator (also a Duration) and subtracts tick_duration until there's <tick_duration left, without ever doing any floating-point math. As you can see, that small error is enough to accumulate into a significant drift for long-running clients, especially in challenging conditions.
 
We could try to replicate the run_fixed_main_schedule logic exactly, or use an integer division + remainder in apply_duration, but I don't know if it's possible to do that without also storing our overstep as a Duration - and anyway I think it's easier and more robust to just copy the LocalTimeline tick and Time<Fixed> overstep any time after RunFixedMainLoop. So, I added a system to do that in SyncedTimelinePlugin and removed InputTimeline::advance_timeline.

### DIFF
--- a/lightyear_core/src/timeline.rs
+++ b/lightyear_core/src/timeline.rs
@@ -48,6 +48,8 @@ pub trait NetworkTimeline: Component<Mutability = Mutable> {
 
     fn overstep(&self) -> Overstep;
 
+    fn set_now(&mut self, now: TickInstant);
+
     fn apply_delta(&mut self, delta: TickDelta);
 
     fn apply_duration(&mut self, duration: Duration, tick_duration: Duration) {
@@ -73,6 +75,10 @@ impl<C: TimelineContext, T: Component<Mutability = Mutable> + DerefMut<Target = 
 
     fn overstep(&self) -> Overstep {
         self.now().overstep
+    }
+
+    fn set_now(&mut self, now: TickInstant) {
+        self.now = now;
     }
 
     fn apply_delta(&mut self, delta: TickDelta) {
--- a/lightyear_sync/src/client.rs
+++ b/lightyear_sync/src/client.rs
@@ -76,7 +76,6 @@ impl Plugin for ClientPlugin {
         // TODO: should the DrivingTimeline be configurable?
         // the client will use the Input timeline as the driving timeline
         app.add_plugins(SyncedTimelinePlugin::<InputTimeline, RemoteTimeline, true>::default());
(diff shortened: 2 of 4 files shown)


Avilad:
Lint failure outside my changes? :flushed:

demos/spaceships/src/server.rs
```diff
@@ -184,6 +192,68 @@ const NAMES: [&str; 35] = [
     "Mr. T",
 ];
 
+#[derive(Resource)]
+struct ServerStallStress {
```
cBournhonesque: Could you add a short comment explaining what this does?
Something like 
"Sleep the thread by `duration` every `interval_ticks` to see how the timeline sync logic handles timeline mismatches"

PR #1368, opened 2025-12-29
Add a warning when clearing messages from a MessageReceiver.

andriyDev:
I wasted several hours today because I was trying to read messages in PreUpdate, which I guess was happening after the clear, resulting in my systems completely missing these messages.

At least with a warning, I could have known my systems were running incorrectly.


cBournhonesque:
Didn't want to trouble you to rebase, merged https://github.com/cBournhonesque/lightyear/pull/1656

PR #1377, opened 2026-01-10
Add reason to disconnected status

ThierryBerger:
I noticed a TODO in the code so I'm here to suggest an implementation.

it can be somewhat long at times but it does the job:

Screenshot 2026-01-10 at 20 40 36


cBournhonesque:
That works, thanks!

PR #1392, opened 2026-01-27
fix: don’t drop PerSenderReplicationState on visibility Lost

rherv:
Observed bug: for entities using `NetworkVisibility`, replication works on initial relevance, but after an entity becomes irrelevant once (visibility transitions to `Lost` and a DESPAWN is sent), calling `ReplicationState::gain_visibility()` later does not reliably re-spawn the entity for that sender.

Root cause: `NetworkVisibilityPlugin::update_network_visibility` dropped the entire `PerSenderReplicationState` entry on `Lost`, which also discards per-sender metadata (e.g. predicted/interpolated/authority/spawned). When the sender entry is recreated later via `entry(...).or_insert_with(...)`, those flags come back in a “blank” state and are not re-applied by visibility helpers.

### Fix

Stop deleting the per-sender entry on `Lost`. Instead, keep the entry and reset it to a safe baseline for future re-gains (e.g. clear `spawned` and move visibility back to a non-visible state), so a subsequent `gain_visibility()` correctly triggers a fresh SPAWN without losing metadata.

### DIFF
--- a/lightyear_replication/src/visibility/immediate.rs
+++ b/lightyear_replication/src/visibility/immediate.rs
@@ -110,13 +110,15 @@ impl NetworkVisibilityPlugin {
                     state.visibility = VisibilityState::Visible;
                 }
 
-                // TODO: is it safe to discard the PerSenderReplicationState information?
-                //  what if we knew that we don't have authority over the entity
-                //  so even if we call `gain_visibility` we want to keep that information?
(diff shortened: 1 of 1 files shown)


cBournhonesque:
I merged https://github.com/cBournhonesque/lightyear/pull/1401/changes which does something similar.
Thanks for the PR!

PR #1404, opened 2026-02-10
fix: pipeline jobs test and format fail

invertedEcho:
Hi, I've noticed the pipeline has been failing for some time now and thought I can help with that.

The tests fail to compile as leafwing is used but the keyboard feature is not enabled.


cBournhonesque:
It seems that the tests pass with your change! 
I tried it myself and merged, thanks for your help

PR #1421, opened 2026-03-07
interpolation: initialize confirmed history when Interpolated is added

Dastari:
## Summary
- initialize `ConfirmedHistory<C>` when `Interpolated` is added to an entity that already has `Confirmed<C>`
- keep the existing `Confirmed<C> -> ConfirmedHistory<C>` insertion path
- add regression tests for both insertion orders

## Problem
`ConfirmedHistory<C>` was only initialized when `Confirmed<C>` was inserted after `Interpolated` was already present.

That left a gap for the opposite insertion order:
- entity already has `Confirmed<C>`
- `Interpolated` is added later

In that case interpolation state could be missing until a later update repaired it implicitly, which is the wrong contract.

## Fix
Add a complementary observer that reacts to `On<Add, Interpolated>` and inserts `ConfirmedHistory<C>` when the entity already has `Confirmed<C>` and does not already have history.

This makes both valid insertion orders behave the same:
- `Interpolated` -> `Confirmed<C>`
- `Confirmed<C>` -> `Interpolated`

## Tests
- inserts history when confirmed is added after interpolated
- inserts history when interpolated is added after confirmed
- does not insert history when no confirmed component exists

## Notes

This fixes a TODO that was already present.

### DIFF
--- a/lightyear_interpolation/src/interpolation_history.rs
+++ b/lightyear_interpolation/src/interpolation_history.rs
@@ -119,6 +119,24 @@ pub(crate) fn insert_confirmed_history<C: Component>(
     }
 }
 
+/// When [`Interpolated`] is inserted on an entity that already has [`Confirmed<C>`], we also
+/// need to ensure [`ConfirmedHistory<C>`] exists.
+///
+/// This complements [`insert_confirmed_history`] so both marker insertion orders are supported:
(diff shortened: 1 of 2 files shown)


cBournhonesque:
Thanks, this makes sense to me!

PR #1425, opened 2026-03-09
Fix missing controlled component for HostClient mode

BrainBacon:
The `Controlled` component was missing on the host when running in host-client mode.

### DIFF
--- a/lightyear_replication/src/host.rs
+++ b/lightyear_replication/src/host.rs
@@ -17,7 +17,7 @@ pub struct HostServerPlugin;
 #[derive(QueryData)]
 struct HostServerQueryData {
     entity: Entity,
-    replicate_like: &'static ReplicateLikeChildren,
+    replicate_like: Option<&'static ReplicateLikeChildren>,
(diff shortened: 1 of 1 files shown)


gudgeirbui:
I'm having the same issue with HostClient mode. I tried the `bevy_enhanced_inputs` example with your fix and, as you say, it's still not working. I think that is because the `Controlled` component is still missing, so the inputs don't get applied:
```
if controlled {
    // add Action entities to the predicted Context
    commands.spawn((
            ActionOf::<Player>::new(entity),
            Action::<Movement>::new(),
            Bindings::spawn(Cardinal::wasd_keys()),
    ));
}
```
(from `client.rs` in `bevy_enhanced_inputs`)



I created an issue here https://github.com/cBournhonesque/lightyear/issues/1434

gudgeirbui:
I was wrong. This does fix the missing `Controlled` component. The issue with `bevy_enhanced_inputs` example is that the `Controlled` component is not detected early enough.

BrainBacon:
Thanks for digging into the example. I'll remove the draft status from this PR then.

PR #1427, opened 2026-03-12
Fix child collider sync

### LINKED ISSUE #1128: lightyear_avian interferes with child collider sync
Avian has a system that syncs Colliders that have no rigid body of their own to the RigidBody in their parent [here](https://github.com/Jondolf/avian/blob/c15549d6f752ca27c83ace3745e9eaa2b314d674/src/collision/collider/collider_transform/plugin.rs#L66).

The default out of the box lightyear_avian setup seems to get in the way of this,  a hierarchy such as:

```

Entity
  - RigidBody::Dynamic
  - LinearVelocity / Other physics components
  - Collider
  - Children
       - Entity
           - Collider
```

If you add linear velocity to the Parent Entity, the Child's global transform is not updated (the collider stays where it is and doesn't move with its parent). The Position / Rotation objects are not updated by avian's systems, but if you manually add back such a system:

```rust
pub struct TestPlugin;
impl Plugin for TestPlugin {
  fn build(&self, app: &mut App) {
    app.add_systems(FixedPostUpdate, update_child_collider_position.in_set(PhysicsSet::Sync));
  }
}

pub fn update_child_collider_position(
    mut collider_query: Query<
        (
            &ColliderTransform,
            &mut Position,
            &mut Rotation,
            &ColliderOf,
        ),
        Without<RigidBody>,
    >,
    rb_query: Query<(&Position, &Rotation), (With<RigidBody>, With<Children>)>,
) {
    for (collider_transform, mut position, mut rotation, collider_of) in &mut collider_query {
        let Ok((rb_pos, rb_rot)) = rb_query.get(collider_of.body) else {
            continue;
        };

        position.0 = rb_pos.0 + rb_rot * collider_transform.translation;
        *rotation = *rb_rot * collider_transform.rotation;
    }
}

```

After adding such a system, the Child collider once again moves with its parent. (This is just a copy paste of Avian's current system, so somewhere in the lightyear_avian setup, its either getting overriden or skipped entirely.

I checked that `ColliderTransformPlugin` was added via `app.is_plugin_added::<ColliderTransformPlugin>()` to ensure that avian was still added this plugin and no feature was missing from my setup.

Hilpogar:
Fixes: #1128

I ran into a similar problem on my project and after a few investigation and trials I noticed that the problem was solved if I remove the `app.register_required_components::<Position, ApplyPosToTransform>();` and `app.register_required_components::<Rotation, ApplyPosToTransform>();` in the `sync_position_to_transform` function. I didn't notice sides effect on my project after this change but I don't have a good enough understanding of lightyear + avian codebase to ensure that this doesn't cause issues elsewhere.

### DIFF
--- a/lightyear_avian/src/plugin.rs
+++ b/lightyear_avian/src/plugin.rs
@@ -357,10 +357,6 @@ impl LightyearAvianPlugin {
             .resource::<PhysicsTransformConfig>()
             .position_to_transform
         {
-            // Make sure that PositionToTransform sync also runs for Interpolated entities
-            app.register_required_components::<Position, ApplyPosToTransform>();
-            app.register_required_components::<Rotation, ApplyPosToTransform>();
-
             // TODO(important): handle this
             // NOTE: we do NOT include this because Position/Rotation might not be added at the same time (for example on the Interpolated entity)
             //  we only want to add Transform if both are added at the same time


cBournhonesque:
I believe these are still needed, as it we still want Pos/Rot to Transform propagation for interpolated entities.
I think child collider issues should have been fixed in the latest commit.

PR #1431, opened 2026-03-16
Add InputMarker only in entities with HasAuthority

### LINKED ISSUE #1430: Rebroadcast issues with BEI >0.23
There is a bug with lightyear_bei caused by the new way to handle mock actions in bevy_enhanced_input. In the new version, ActionMock seems to be added in all input entities, even the ones that are rebroadcasting inputs from other clients. Because of that, those message are overflowing the terminal as long as there is more than one client connected :

```
2026-03-16T15:50:26.413620Z ERROR lightyear_inputs::client: received input message for unrecognized entity entity=PLACEHOLDER target_data.states=BEIStateSequence { start_state: ActionsSnapshot { state: None, value: Axis2D(Vec2(0.0, 0.0)), time: ActionTime { elapsed_secs: 0.0, fired_secs: 0.0 }, events: ActionEvents(0) }, diffs: [SameAsPrecedent, SameAsPrecedent, SameAsPrecedent, SameAsPrecedent] } end_tick=Tick(4302)
2026-03-16T15:50:26.413629Z ERROR lightyear_inputs::client: received input message for unrecognized entity entity=1279v0 target_data.states=BEIStateSequence { start_state: ActionsSnapshot { state: None, value: Axis3D(Vec3(0.0, 0.0, 0.0)), time: ActionTime { elapsed_secs: 0.0, fired_secs: 0.0 }, events: ActionEvents(0) }, diffs: [SameAsPrecedent, SameAsPrecedent, SameAsPrecedent, SameAsPrecedent] } end_tick=Tick(4302)
```

In this example, entity 1279v0 was the input entity of the other client

Hilpogar:
Fixes: #1430

Since `ActionMock` is added in all input entities since BEI 0.23, we should filter to only add `InputMarker` in the input entities with `HasAuthority`.

### DIFF
--- a/lightyear_inputs_bei/src/marker.rs
+++ b/lightyear_inputs_bei/src/marker.rs
@@ -3,6 +3,7 @@
 use bevy_ecs::prelude::*;
 use bevy_ecs::relationship::Relationship;
 use bevy_enhanced_input::prelude::*;
+use lightyear_replication::prelude::HasAuthority;
 
 /// Marker component that indicates that the entity is actively listening for physical user inputs.
 ///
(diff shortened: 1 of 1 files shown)


cBournhonesque:
I'm a bit hesitant to use `HasAuthority` as I will be replacing lightyear's replication logic with replicon soon, and that will probably remove the concept of authority (as it wouldn't be possible to implement in replicon for now).

Maybe we could use `Controlled`, but that has its own set of issues.
Anyway thanks for flagging this, I wasn't aware that ActionMock was now added to all input entities.

PR #1433, opened 2026-03-19
Update avian2d/avian3d from 0.5 to 0.6

baszalmstra:
Update avian to 0.6. Adapts `update_collision_layers` for immutable `CollisionLayers`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -225,8 +225,8 @@ leafwing-input-manager = { version = "0.20", default-features = false, features 
 ] }
 
 # physics
-avian2d = { version = "0.5", default-features = false }
-avian3d = { version = "0.5", default-features = false }
+avian2d = { version = "0.6", default-features = false }
+avian3d = { version = "0.6", default-features = false }
 
 # gui debug ui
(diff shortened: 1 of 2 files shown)


baszalmstra:
Im having a hard time getting the move-and-slide example to work with lightyear. I think it is because the example modifies the Transform directly. It then uses `CustomPositionIntegration` to overwrite the position integration behavior using velocities. However, this conflicts with what lightyear does. 

Help would be appreciated. :)

baszalmstra:
After some tinkering I figured out there is no need to integrate `CustomPositionIntegration` in the plugin. :) This PR is ready!

andriyDev:
It looks like the `fps` example is broken, complaining about ambiguities between `compute_hit_lag_compensation` and `clear_moved_proxies`/`update_solver_body_aabbs<Collider>`. I'm guessing this would also trigger in a real app.

CaptainDerank:
The `avian_3d_character` example is also broken. The Player and the Box (if you change the Y Position up manually) are falling through the Floor.

PR #1438, opened 2026-03-26
Fix Dropped Inputs and Collapsed JustPressed Events

AdamWhitehurst:
Two commits targetting different issues:
- Predicted entities inputs were being dropped. Server was using `get` instead of `get_predict` for input snapshots. This still seems to be consistent with the comment that claims to consider "that the player will keep playing the last action they played" when no snapshot present because afaict the `get_predict` will `get_last` so same behavior as comment
- The server was not recognizing `just_pressed` but recognizing `pressed`. `ActionDiff` has no `JustPressed` variant, so every button press arrived as `Pressed` on the server, and the old `from_snapshot`  raw-cloned it as-is. Instead of raw-cloning the snapshot, `from_snapshot_transitions` compares the previous server state against the new snapshot and calls `press()`/`release()`, which produces `JustPressed`/`JustReleased` when a transition is detected.

### DIFF
--- a/lightyear_inputs/src/input_message.rs
+++ b/lightyear_inputs/src/input_message.rs
@@ -235,6 +235,24 @@ pub trait ActionStateSequence:
     /// Modify the given state to reflect the given snapshot.
     fn from_snapshot<'w>(state: StateMutItemInner<'w, Self>, snapshot: &Self::Snapshot);
 
+    /// Apply a snapshot to the state while preserving button transition semantics.
+    ///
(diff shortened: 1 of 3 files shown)


cBournhonesque:
I will have to be very careful when reviewing this so it might take a while; I iterated a lot on the correct handling of inputs so I'm not convinced that your changes are correct.

cBournhonesque:
Ok this makes sense to me!

PR #1448, opened 2026-04-13
fix: register Transform as required for Collider for avian 0.6 compatibility

### LINKED ISSUE #1253: Collision issue with avian
Using this commit and going to this folder: [cBournhonesque/lightyear@b5392fc/examples/deterministic_replication](https://github.com/cBournhonesque/lightyear/tree/b5392fc0a3b31f624dfd48b2eda5ea4665474a58/examples/deterministic_replication)

If you run (in 3 different processes)

cargo run --no-default-features --features=server,netcode,udp -- server
cargo run -- client -c 1
cargo run -- client -c 2
And start moving the player entity from client 2 towards the ball, then 80% of the time things are working and 20% of the time
on client 1's screen the player will not collide with anything. It will pass through the ball and the wall and go to outer space.

I've checked the components on the entity and everything seems to be correct.

The non-deterministic aspect makes me think of a system ordering issue, but i'm not sure

Avian issue: https://github.com/Jondolf/avian/issues/863

stevesloan:
Builds on #1433 (avian 0.5 → 0.6 upgrade).

## Summary

Avian 0.6's `ColliderOf::on_insert` hook requires `GlobalTransform` to establish the
`ColliderOf → RigidBodyColliders` relationship. When `PhysicsTransformPlugin` is disabled
(as lightyear requires), `GlobalTransform` was missing, causing `ColliderOf::on_insert` to
return early. This left `RigidBodyColliders` empty, so `update_solver_body_aabbs` skipped
the entity and its AABB stayed frozen at the spawn position. The broad phase never detected
collisions and objects fell through each other.

This worked in avian 0.5 because the old sweep-and-prune broad phase rebuilt from scratch
every frame and didn't depend on `RigidBodyColliders`.

The fix registers `Transform` as a required component for `Collider` (not `Position`/`Rotation`,
which would cause issues with Interpolated entities). This ensures `GlobalTransform` is present
when avian's hooks run.

## Related issues

- Fixes #1253 (collision issue with avian)
- Partially addresses #1128 (child collider sync — fixes the `ColliderOf` relationship setup,
  but the `ApplyPosToTransform` interference is a separate issue)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -225,8 +225,8 @@ leafwing-input-manager = { version = "0.20", default-features = false, features 
 ] }
 
 # physics
-avian2d = { version = "0.5", default-features = false }
-avian3d = { version = "0.5", default-features = false }
+avian2d = { version = "0.6", default-features = false }
+avian3d = { version = "0.6", default-features = false }
 
 # gui debug ui
 bevy-inspector-egui = { version = "0.36", default-features = false, features = [
--- a/lightyear_avian/src/lag_compensation/history.rs
+++ b/lightyear_avian/src/lag_compensation/history.rs
@@ -132,25 +132,24 @@ fn spawn_broad_phase_aabb_envelope(
 
 /// Update the collision layers of the child AabbEnvelopeHolder to match the parent
 fn update_collision_layers(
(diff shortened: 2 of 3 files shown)


cBournhonesque:
Great find, thanks!

Issue #1450, opened 2026-04-20
Interpolated component never converges on `Confirmed` once `ConfirmedHistory` shrinks to one keyframe


monsterrdev:
I noticed some subtle visual desync between two client connections while validating my initial configuration on a fresh bevy+lightyear project. I'm just building some initial familiarity with the framework at this point, and a lot of what I've been doing has been agent driven so it is possible I may have misconfigured something or misunderstood expected behaviour.

## Summary

When a server-replicated, interpolated entity stops changing, its interpolated
value on the client never settles on the authoritative `Confirmed` value. The
client's `ConfirmedHistory<C>` buffer drains to `len == 1`, and from then on
`ConfirmedHistory::interpolate` returns `None` — the interpolated component
stays frozen at whatever blended value it held just before the second-to-last
keyframe was popped. Non-zero drift persists indefinitely.

## Environment

- `lightyear = 0.26.4` (also reproduced on current `main` — same code path)
- `bevy = 0.18`
- Features: `udp`, `netcode`
- OS: macOS (Darwin 24.5.0); not platform-specific — pure ECS/logic bug.

## Steps to reproduce

Minimal standalone repro is attached (see `src/bin/server.rs`,
`src/bin/client.rs`, `src/lib.rs`, `Cargo.toml`). Summary:

1. Server spawns one entity with `Position(Vec2)`, `Replicate::to_clients(All)`,
   `InterpolationTarget::to_clients(All)`. A `ReplicationSender` is attached
   per link with `SendUpdatesMode::SinceLastAck` and a 40 ms send interval.
2. Once a client links, the server ticks `Position.x += 0.05` for 320 `FixedUpdate`
   steps (~5 s at 64 Hz), then stops mutating the component. No further
   updates are sent (the value isn't changing, and prior keyframes were acked).
3. Client runs `ClientPlugins { tick_duration: 1/64 s }` plus an
   `InterpolationConfig::default()`, and logs once per second:
   - `pos`: the currently-displayed interpolated `Position`
   - `confirmed`: `Confirmed<Position>` (server's last replicated value)
   - `len` / `newest`: keyframe count and newest tick in `ConfirmedHistory`
   - `drift`: `(pos - confirmed).length()`

Run:

```
cargo run --bin repro_server    # in one terminal
cargo run --bin repro_client    # in another
```

## Expected behavior

After the server stops mutating the component, the client's interpolated
`Position` should converge on `Confirmed` (drift → 0) — either by keeping the
[...]

PR #1451, opened 2026-04-20
fix(interpolation): converge on latest confirmed value under packet loss

monsterrdev:
This is an attempt to resolve the issue described here https://github.com/cBournhonesque/lightyear/issues/1450

I went through a few iterations with the assistance of a coding agent, the latest commit still displays jitter when simulating a poor connection (400ms+), but solves the issue of drift between what clients see and what the server has as state. I suspect there is room for improvement, but I don't have enough context to clearly identify it.

Agent generated context below.

## Summary

Fixes persistent drift and jitter on remote `Interpolated` entities when updates stop arriving or are lost. Interpolation now converges cleanly on the latest confirmed value during true idle, clamps on the freshest bracketing pair during short loss gaps, and no longer snaps mid-blend when the buffer transiently drains.

## Problem

`ConfirmedHistory<C>::interpolate` required two keyframes and returned `None` otherwise. `update_confirmed_history`'s rebase branch was the only path that could advance a single-keyframe buffer to the current tick, and it gated its component write on `!present`. Result: after the server stopped sending updates, the interpolated component stayed pinned to its last blended value — slightly behind `newest` — and never converged. Reproduced in a minimal standalone app; persistent ~0.0006 unit drift on localhost, larger under link conditioning.

## Iteration history

I tried four algorithmic variants; each fixed the previous one's failure mode and surfaced a new one. The branch is kept as separate commits for reviewability; happy to squash if preferred.

**1. `32d4b8c5` — Snap on single keyframe.** Made `interpolate()` snap to `start()` when `len == 1` so the idle-convergence case was covered without relying on the rebase's write. Fixed the drift but introduced visible jitter on remote entities during active motion at >200ms RTT: any transient drain to `len == 1` mid-blend produced an instantaneous snap to the newest anchor.

**2. `de25275b` — Time-gated rebase writes every tick.** Reverted `interpolate()` to the `len >= 2` contract; moved the write into the idle rebase path (no longer `!present`-gated), firing every tick the `len == 1 && current - newest >= send_interval_delta_tick` gate held. Smooth under true idle, but the 4-tick gate isn't idle-specific — it also fires on loss gaps during active motion. Each packet-loss event produced one stale write followed by a catch-up blend on the next arrival. Subtle at 2% loss, rough at 10%.

**3. `5258e8d0` — Structural: maintain bracketing pair.** Rewrote `update_confirmed_history` to keep `[behind, newest]` while updates flow, clamped `fraction` in `interpolate()` to `[0, 1]` to prevent extrapolation past `newest`, and re-added snap-on-`len == 1`. Eliminated the idle jerk but regressed mid-motion: the drain used `while len >= 2 && get_nth(1) <= current: pop`, which collapsed to `len == 1` on every gap, giving more visible jumps per gap than the time-gated version.

**4. `adfb183a` — Smart drain + clamp.** Tightened the drain to `while len >= 3 && get_nth(1) <= current: pop` so the buffer never drops below 2 during active replication. The idle rebase now fires from `len >= 2` (the drain doesn't reduce below 2 on its own), collapsing to a single retagged anchor at the current tick and writing the component directly. Removed snap-on-`len == 1` from `interpolate()` since the rebase is the only path that can produce `len == 1`. Clamp retained: bounded jerk on direction reversals is preferable to extrapolation overshoot when the server path diverges from a linear extrapolation.

**5. `8edf6945` — Restore context comments.** Pure docs. Restored three comments/TODOs that were still load-bearing: the "exact start tick" TODO on the idle rebase (the proxy-vs-real-tick problem remains open), the `X...H1...H2` explanation for the `interpolation_tick < start_tick` guard in `interpolate()`, and a design-decision note at the clamp site describing why clamp was chosen over extrapolation.

## Final design

`interpolate()` is now a pure geometric operation:
- `len == 0` → `None`
- `len == 1` → `None` (idle rebase path is responsible for the component write)
- `len >= 2` → clamped blend between `start()` and `end()`

`update_confirmed_history` maintains `[behind, newest]` during active replication (smart drain) and collapses to a single rebased anchor at the current tick when `current - newest >= send_interval_delta_tick` (idle rebase). The rebase writes the component directly.

Extrapolation past `end_tick` is intentionally not implemented. Clamp bounds the jerk on direction reversals to `|newest - behind|`; extrapolation would look smoother for continuous motion but amplify mismatch when the server path diverges. The TODO at the previous extrapolation site is converted into an inline design-decision comment tied to velocity-aware interpolation with reconciliation as a future direction.

## Test plan

- [x] Unit tests in `lightyear_interpolation` (`ConfirmedHistory::interpolate`) updated and passing.
- [x] Standalone minimal repro (originally used to isolate the drift bug): drift eliminated.
- [x] In-game testing with a downstream Bevy project using `LinkConditioner` presets:
  - `good` (40ms, 6ms jitter, 0.2% loss) — smooth.
  - `average` (100ms, 15ms jitter, 2% loss) — smooth with occasional jitter at ~25s cadence (matches the probability of 2 consecutive drops).
  - `poor` (200ms, 30ms jitter, 10% loss) — visibly lossy; 2-drop events occur ~every 2s and can't be hidden by interpolation alone. Raising `InterpolationConfig::send_interval_ratio` from `1.7` to `3.0` materially improves this by buffering an extra keyframe.

### DIFF
--- a/lightyear_interpolation/src/interpolate.rs
+++ b/lightyear_interpolation/src/interpolate.rs
@@ -24,21 +24,27 @@ pub fn interpolation_fraction(start: Tick, end: Tick, current: Tick, overstep: f
     ((current - start) as f32 + overstep) / (end - start) as f32
 }
 
-/// Update the ConfirmedHistory so that interpolation can simply interpolate between the last 2 updates.
+/// Maintain the ConfirmedHistory so that `interpolate()` always sees the right anchors.
+///
+/// Goal: keep the history at `[behind, newest]` while updates flow, where `behind` is the
+/// most recent keyframe at or before the current interpolation tick. Under packet loss this
+/// means `interpolate()` blends between the freshest pair of confirmed values that bracket
+/// the current tick (or clamps at `newest` when we momentarily run out of forward anchors),
+/// instead of falling back to a single-keyframe rebase that snaps on every gap.
+///
+/// We only collapse to a single keyframe when the entity has been idle long enough that the
+/// next update is genuinely a fresh start, in which case we rebase the lone keyframe to the
+/// current tick so a future update interpolates from "now" rather than from a stale tick.
 pub(crate) fn update_confirmed_history<C: Component + Clone>(
     // TODO: handle multiple interpolation timelines
     // TODO: exclude host-server
     interpolation: Single<&InterpolationTimeline, With<IsSynced<InterpolationTimeline>>>,
     tick_duration: Res<TickDuration>,
-    // we don't insert the component immediately, instead we wait for:
-    // - either 2 updates, so that we can interpolate between them
-    // - or enough time has passed since the initial update
     mut query: Query<(Entity, &mut ConfirmedHistory<C>, Has<C>)>,
     mut commands: Commands,
 ) {
     let timeline = interpolation.into_inner();
 
-    // how many ticks between each interpolation
     let send_interval_delta_tick = (SEND_INTERVAL_TICK_FACTOR
         * timeline.remote_send_interval.as_secs_f32()
         / tick_duration.as_secs_f32())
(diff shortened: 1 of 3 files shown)


cBournhonesque:
Thanks that's a great catch; i thought my current logic was always trying to keep at least 2 frames in the history in most cases to avoid this issue, but apparently not

Issue #1446, opened 2026-04-06
Connected transition for crossbeam transport


utgarda:
_Originally posted by @utgarda in https://github.com/cBournhonesque/lightyear/discussions/1432_

## Problem

`CrossbeamPlugin` is missing the `Linked` → `Connected` state transition, making crossbeam unusable as a real in-process transport outside the internal test stepper. After firing the `Connect` trigger, `LinkStart` inserts `Linked` but nothing advances the entity to `Connected` — the client waits forever.

The same gap exists on the server side: `LinkOf` mirror entities never get `Connected` or `ClientOf`.

## Root cause

Every other transport has an observer that promotes `Linked` to `Connected`:

| Transport | Linked → Connected observer |
|---|---|
| `RawClient` | `on_linked` in `lightyear_raw_connection/src/client.rs` |
| `NetcodeClient` | `receive` system in `lightyear_netcode/src/client_plugin.rs` |
| `SteamClientIo` | `on_linked` in `lightyear_steam/src/client.rs` |
| **CrossbeamIo** | **missing** |

## Workaround

Our game maintains a local patch of `lightyear_crossbeam` with the missing observers and falls back to a local UDP server for integration tests that need to exercise the real network stack.

## Fix

Two observers added to `CrossbeamPlugin`, following the `lightyear_raw_connection` pattern:

- `on_client_linked` — inserts `Connected`, `LocalId`, `RemoteId` when `Linked` is added on a `CrossbeamIo` + `Client` entity
- `on_server_client_linked` — inserts `Connected`, `LocalId`, `RemoteId`, `ClientOf` when `Linked` is added on a `CrossbeamIo` + `LinkOf` entity

Also improved the `send` system to handle disconnected channels gracefully (peer dropped / shutdown) instead of returning an error.

Tests cover client connection, server mirror connection, and send-after-disconnect.

PR #1459, opened 2026-04-28
crossbeam: graceful Disconnected/Full handling + transport-pairing docs

utgarda:
## Summary

Three changes that make `lightyear_crossbeam` more robust as a transport, without changing its layering responsibilities:

### 1. `send` system: graceful disconnect + `Full` re-queue

The previous `send` used `try_for_each(drain)` and propagated `try_send` errors out of the closure. The first failure surfaced as a system-level error and short-circuited *all* sends across all entities for that frame, even though the per-entity links are independent.

New behavior, per-entity:
- `Disconnected` → log at `trace!` (not an error during shutdown), clear that entity's send queue, continue with other entities.
- `Full` → log at `error!` (this transport assumes unbounded channels — see (2)), `push_front` the failed payload so FIFO is preserved across the still-queued payloads, continue with other entities.
- `Ok` → keep draining.

Signature changed from `-> Result` to `()` — there are no longer any errors that callers can usefully react to.

### 2. `LinkSender::push_front` (small `lightyear_link` addition)

The `Full` re-queue path puts a `pop`'d payload back at the front of the deque. Added `pub fn push_front(&mut self, value: SendPayload)` to `LinkSender` — one line wrapping `VecDeque::push_front`. Non-breaking, reusable for any future transport that needs the same backpressure pattern.

### 3. Module docs: connection-layer pairing

Added a "Connection layer" section to `lightyear_crossbeam`'s module docs explaining that `CrossbeamPlugin` is a pure transport (it raises `Linked` but does not drive `Connected`), and pointing users at:
- `lightyear_raw_connection::client::RawConnectionPlugin` / `server::RawConnectionPlugin` (with `RawClient` / `RawServer` markers) for handshake-less use.
- `lightyear_netcode` for authenticated use.

Documented `CrossbeamIo::new` as requiring an unbounded sender, with `new_pair` flagged as the canonical configuration.

## Relationship to #1446

#1446 originally asked for `Linked → Connected` observers directly inside `CrossbeamPlugin`, mirroring `lightyear_steam` / `lightyear_netcode`. An earlier draft of this PR took that approach but it was reverted: `lightyear_raw_connection` already provides the same mechanism via its `RawClient` / `RawServer` opt-in markers, and putting observers in the transport layer broke composition with `lightyear_netcode` (whose `connect` system expects `Connecting` to still be present, which `Connected::on_add` would have removed). The doc additions in (3) point users at the existing solution.

## Tests

```bash
cargo test -p lightyear_crossbeam --all-features
cargo clippy -p lightyear_crossbeam -p lightyear_link --all-features --tests -- -D warnings
cargo clippy -p lightyear_crossbeam -p lightyear_link --no-default-features --tests -- -D warnings
```

Three new transport-only tests in `lightyear_crossbeam`:
- `send_after_peer_disconnect_does_not_panic` — drops the peer side, queues two payloads, runs one frame, asserts the send queue ends empty.
- `round_trip_send_receive` — paired `(client_io, server_io)`, three payloads, asserts receive-side FIFO over two frames (`PostUpdate` send → next frame's `PreUpdate` receive).
- `send_with_bounded_channel_requeues_on_full` — wires a `bounded(1)` sender via `CrossbeamIo::new`, queues `["first", "second", "third"]`, asserts the queue ends `["second", "third"]` so FIFO is preserved across the `Full` re-queue.

## Compatibility

- No breaking API changes; `LinkSender::push_front` is purely additive.
- No new dependencies.
- `#![no_std]` preserved.

The branch has been validated end-to-end in a downstream consumer project (multiplayer Bevy game) where it fixes a pre-Countdown crash on local games when the server side restarts mid-frame.

### DIFF
--- a/lightyear_crossbeam/Cargo.toml
+++ b/lightyear_crossbeam/Cargo.toml
@@ -14,7 +14,6 @@ test_utils = ["lightyear_core/test_utils"]
 
 [dependencies]
 lightyear_core.workspace = true
-lightyear_connection = { workspace = true, features = ["client", "server"] }
 lightyear_link.workspace = true
 
 # bevy
--- a/lightyear_crossbeam/Cargo.toml
+++ b/lightyear_crossbeam/Cargo.toml
@@ -28,6 +28,12 @@ tracing.workspace = true
 # serde
 bytes.workspace = true
 
+[dev-dependencies]
+# Used by integration tests that exercise the documented pairing with
+# `RawConnectionPlugin` (server-side mirror flow).
+lightyear_raw_connection = { workspace = true, features = ["client", "server"] }
+lightyear_connection = { workspace = true, features = ["client", "server"] }
+
 [lints]
 workspace = true
 
--- a/lightyear_crossbeam/Cargo.toml
+++ b/lightyear_crossbeam/Cargo.toml
@@ -14,6 +14,7 @@ test_utils = ["lightyear_core/test_utils"]
 
 [dependencies]
 lightyear_core.workspace = true
+lightyear_connection = { workspace = true, features = ["client", "server"] }
 lightyear_link.workspace = true
 
 # bevy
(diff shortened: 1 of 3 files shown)


cBournhonesque:
That works for me, thanks for the extra documentation and the fix!

PR #1465, opened 2026-05-06
Fix Tick: Ord transitivity violation corrupting BTreeMap<Tick, _>

3264studios:
## Summary

`Tick`'s `Ord` impl uses signed wrap-aware semantics, which violates the transitivity contract: `a < b && b < c` does not always imply `a < c`. This silently corrupts any `BTreeMap<Tick, _>` whose entries span more than 32 768 ticks (~13.65 min at 40 Hz).

`lightyear_replication/src/delta.rs` has two such maps — `DeltaComponentHistory.buffer: BTreeMap<Tick, C>` and `DeltaManager.data: BTreeMap<Tick, PerTickData>` — both populated continuously during a session, both vulnerable.

## Reproducer (verifiable on pristine `main`)

Drop these into `lightyear_core::tick::tests` and run `cargo test -p lightyear_core`:

```rust
#[test]
fn upstream_tick_ord_transitivity() {
    let t1 = Tick(0);
    let t2 = Tick(32_768);
    let t3 = Tick(33_000);
    assert!(t1 < t2);  // wrapping_diff = -32_768 -> Greater -> fails
    assert!(t2 < t3);
    assert!(t1 < t3);  // wrapping_diff = -32_536 -> Greater -> fails
}

#[test]
fn upstream_tick_ord_btreemap_consistency() {
    use std::collections::BTreeMap;
    let mut m: BTreeMap<Tick, u32> = BTreeMap::new();
    for t in [0u16, 16_000, 32_768, 33_000, 49_000] {
        m.insert(Tick(t), t as u32);
    }
    for t in [0u16, 16_000, 32_768, 33_000, 49_000] {
        assert!(m.get(&Tick(t)).is_some(), "BTreeMap.get failed for tick {t}");
    }
}
```

Verified on `main@013c51fa`:
```
test tick::tests::upstream_tick_ord_transitivity ... FAILED
  assertion failed: t1 < t2
test tick::tests::upstream_tick_ord_btreemap_consistency ... FAILED
  BTreeMap.get failed for tick 16000
```

## Fix

**Type-system foundation (commit 1):** Remove `Ord` / `PartialOrd` from the `wrapping_id!` macro. Give `Tick` a *total* numeric (`u16`) `Ord` that satisfies transitivity. Add explicit wrap-aware methods at the `Tick` site:

- `Tick::is_newer_than(self, other) -> bool`
- `Tick::is_older_than(self, other) -> bool`
- `Tick::wrapping_cmp(&self, other) -> core::cmp::Ordering`

`MessageId`, `PacketId`, and `PingId` retain wrap-aware `Ord` at their definition sites — transport channels' in-order delivery depends on it. Those types carry a latent BTreeMap-corruption risk under very high message rates, but this PR limits scope to `Tick` (the type that actually appears as a `BTreeMap` key in delta paths today).

**Site migration (commit 2):** Every site that previously relied on `Tick: Ord` for sequence semantics now uses the explicit method. Sites unchanged: those already using `Sub` returning `i16` (e.g. `tick - other > delta`), equality checks, and numeric tick deltas.

**Pre-existing fmt drift (commit 3):** Three files (`lag_compensation/history.rs`, `receive.rs`, `visibility/room.rs`) failed `cargo fmt --check` on `main` before this PR; the Lint job has been red since 2026-04-26 because of them. Folded the mechanical fixes here so this PR can go green on Lint. Drop this commit during squash-merge if a separate fmt PR is preferred.

## Tests

Four new tests in `lightyear_core::tick`:
- `tick_ord_transitivity` — `t1 < t2 < t3` across the i16 boundary
- `tick_ord_btreemap_consistency` — round-trip insert / lookup across the wrap
- `tick_ord_is_total` — total-order spot checks
- `tick_wrap_aware_methods` — `is_newer_than` / `is_older_than` semantics

```
$ cargo test -p lightyear_core tick_
running 6 tests
... 6 passed; 0 failed
```

`cargo build --workspace`: clean. `cargo test -p lightyear_inputs` (14/14) and `cargo test -p lightyear_replication` (19/19): all pass.

## Notes

- This was discovered while investigating an unrelated production issue. The downstream gameplay symptom turned out to have a different root cause, but the latent correctness bug here is real, reproducible on pristine `main`, and worth fixing on its own merits.
- A follow-up could give `MessageId` / `PacketId` / `PingId` the same split-treatment if any of them ever land in a `BTreeMap` keyed structure.

### DIFF
--- a/lightyear_avian/src/lag_compensation/history.rs
+++ b/lightyear_avian/src/lag_compensation/history.rs
@@ -139,20 +139,18 @@ fn update_collision_layers(
     >,
     mut commands: Commands,
 ) {
-    parent_query
-        .iter()
-        .for_each(|(parent, layers, children)| {
-            for child in children.iter() {
-                if child_query.get(child).is_ok() {
-                    commands.entity(child).insert(*layers);
-                    trace!(
-                        ?child,
-                        ?parent,
-                        "Adding layers {layers:?} on lag compensation child collider"
-                    );
-                }
+    parent_query.iter().for_each(|(parent, layers, children)| {
+        for child in children.iter() {
+            if child_query.get(child).is_ok() {
+                commands.entity(child).insert(*layers);
+                trace!(
+                    ?child,
+                    ?parent,
+                    "Adding layers {layers:?} on lag compensation child collider"
+                );
             }
-        });
+        }
+    });
 }
 
 /// For each lag-compensated collider, store every tick a copy of the
--- a/lightyear_core/src/history_buffer.rs
+++ b/lightyear_core/src/history_buffer.rs
@@ -100,7 +100,7 @@ impl<R> HistoryBuffer<R> {
     /// Used to efficiently get the previous value when doing correction.
     pub fn second_most_recent(&self, tick: Tick) -> Option<&R> {
         let (most_recent_tick, most_recent) = self.most_recent()?;
-        if *most_recent_tick < tick {
+        if most_recent_tick.is_older_than(tick) {
             return most_recent.into();
         }
         let len = self.buffer.len();
@@ -115,7 +115,7 @@ impl<R> HistoryBuffer<R> {
         // find first idx `partition` such that self.buffer[partition].0 > tick
         let partition = self
             .buffer
-            .partition_point(|(buffer_tick, _)| *buffer_tick <= tick);
+            .partition_point(|(buffer_tick, _)| buffer_tick.wrapping_cmp(&tick).is_le());
         if partition == 0 {
             return None;
         }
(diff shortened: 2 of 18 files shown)


3264studios:
Heads-up on the third commit: `style: fix pre-existing rustfmt drift on main` is mechanical `cargo fmt` on three files (`lag_compensation/history.rs`, `receive.rs`, `visibility/room.rs`). All three failed `cargo fmt --check` on `main@013c51fa` before this PR was opened — the Lint job has been red on `main` since 2026-04-26 because of them. I folded the fixes here opportunistically so this PR's Lint job can go green; if you'd prefer them in a separate PR, dropping this commit during squash-merge is fine and won't affect the Tick correctness fix.

The `Test` job failure is unrelated to this PR: `bevy_enhanced_input::action::ActionMock` became private in v0.24.4, breaking `lightyear_tests/src/host_server/input/bei.rs:4`. Same failure on `main`.

cBournhonesque:
Thanks for the PR!
I actually recently switched Tick from u16 to u32 to avoid these tick-wrapping headaches. 
I don't think the complexity is worth the extra 2 bytes.

Sorry that your PR doesn't apply anymore, the fix looked good to me.

PR #1468, opened 2026-05-10
Fix incorrect resource removal during rollback

### LINKED ISSUE #1467: Rollback resources incorrectly removed after sync+rollback
Here's the scenario:
1. Register resource `Foo` to be rollbacked: `app.add_resource_rollback::<Foo>()`
1.a An empty `ResourceHistory<Foo>` is added
1.b An observer is spawned that captures `Foo` whenever `PredictionManager` is added.
2. Add resource `Foo`: `app.init_resource::<Foo>()`.
3. Add lightyear's client plugin
3.a. Local tick starts at 0.
4. Spawn `Client` and `PredictionManager` component in order to connect to a server.
4.a. By now, the local tick is small (say, tick 5).
4.b. `PredictionManager` causes observer in step 1.b to fire and write `Foo` into history: The whole history is now `{5: value}`.
5. Lightyear's netcode handshake completes with a server whose current tick is 100.
5.a. As part of joining the server's timeline, a sync event fires with `tick_delta = 95`.
5.b. `handle_tick_event_resource_history` shifts every history entry forward by `tick_delta`.
5.c. `Foo`'s history becomes `{100: value}`
6. Confirmed state for some other entity (but not for `Foo` yet) from the server arrives for tick 99
7. Lightyear initiates a rollback with `rollback_target_tick = 99`.
7.a. `prepare_rollback_resource::<Foo>` calls `history.clear_except_tick(99)`. The only entry is at tick 100 so `clear_except_tick` returns `None`.
**7.b. `prepare_rollback_resource` *interprets* `None` as "`Foo` did not exist at tick 99" and calls `commands.remove_resource::<Foo>()`. The interpretation is wrong. `Foo` has been in the world since step 2. Lightyear just has not captured a value at-or-before tick 99 because the only capture (step 4) happened at a tick that has since been shifted past 99 by the sync.**
8. In the next `FixedUpdate`, any system reading `Res<Foo>` (non-`Option`) fails Bevy's system-param validation resulting in a panic.

I've encountered this problem while setting `LightyearAvianPlugin::rollback_resources` to `true`. The setting marks some of avian's resources (e.g. `ContactGraph`)  to be able to rollback and then this scenario happens causing avian to panic.

nick-e:
`prepare_rollback_resource` treated `clear_except_tick` returning `None` as "the value was absent at this tick" and removed the resource. But `clear_except_tick` also returns `None` when the history buffer is non-empty but every entry is after the requested tick.  This is a typical scenario when a client joins a server and performs its first rollback.

Removing the resource may cause bevy to panic when trying to run a system that requires access to the resource in the next tick. This is an issue with `LightyearAvianPlugin::rollback_resources` which marks some avian resources has rollbackable.

The fix is to leave the resource alone if there is no value at or before the given rollback tick.

Fixes #1467

### DIFF
--- a/lightyear_prediction/src/rollback.rs
+++ b/lightyear_prediction/src/rollback.rs
@@ -629,10 +629,10 @@ pub(crate) fn prepare_rollback<C: SyncComponent>(
                     "Rollback to the value stored in the PredictionHistory"
                 );
                 (
-                    original_predicted_value.as_ref().and_then(|v| match v {
-                        HistoryState::Updated(v) => Some(v),
+                    match original_predicted_value.as_ref() {
+                        Some(HistoryState::Updated(v)) => Some(v),
                         _ => None,
-                    }),
+                    },
                     true,
                 )
             }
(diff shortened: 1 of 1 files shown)


cBournhonesque:
Thanks, that's a good catch! I had already fixed this while migrating to replicon, so I will close this.

Issue #1402, opened 2026-02-09
Default `no_input_delay` config doesn't reliably deliver inputs on localhost due to sync error margin tolerance


nuzzles:
# Issue

The default `no_input_delay()` config combined with the default `SyncConfig` (error_margin=1.0) is subtly broken on localhost:

- Sync objective = remote + RTT/2 + jitter_margin ≈ remote + 0.075 ticks
- The guard (`if obj < remote`) does NOT trigger because `remote + 0.075 > remote`
- Error margin of 1.0 tick allows the client to drift to `remote - 0.925`
- `end_tick = client_tick ≈ server_tick - 1`: server consistently misses inputs
- `update_action_state` silently keeps stale state (no error, no warning)

TLDR: The issue is that the sync objective guard (line 289-290 of input.rs) uses `obj < remote` but should arguably be `obj < remote + 1` to ensure inputs always arrive on time, regardless of error margin tolerance. Without this, the `no_input_delay()` preset doesn't reliably deliver inputs even on `localhost`.

nuzzles:
Additional context: I was using a 15Hz tick rate.

**The problem**: Inputs are received in PreUpdate but consumed in FixedPreUpdate - which is always at least 1 tick later (after FixedFirst advances the tick). The sync objective doesn't account for this processing gap. 

On localhost:

- obj ≈ remote + 0.075 (RTT/2 ≈ 0, jitter_margin ≈ 5ms)
- Guard doesn't trigger (0.075 > 0)
- error_margin = 1.0 allows drift to remote - 0.925
- end_tick = floor(remote - 0.925) = remote - 1
- Server reads at remote + 1 (after FixedFirst advance) - misses by 2 ticks

Proposed fix - change the guard to always target at least 1 tick ahead:
```
let mut obj = remote + network_delay + jitter_margin - input_delay;
// The server consumes inputs in FixedPreUpdate, which runs 1 tick after
// the tick at which the message was received in PreUpdate. The client must
// run far enough ahead to cover this processing gap.
if obj < remote + TickDelta::from_i16(1) {
    obj = remote + TickDelta::from_i16(1)
}
```

This still allows occasional single-tick misses at the boundary, but it changes the behavior from consistently missing every tick to occasionally missing when drift maxes out

mmannerm:
▎ Adding a reproduction and a separate failure mode that I think is worth folding into the same fix.

[...]

PR #1470, opened 2026-05-13
fix(sync): raise sync_objective floor to remote + 1 (#1402)

mmannerm:
Debugging desync issues on my local machine setup running the server in WSL2, and clients in Windows 11 on same machine, I came up with the issue #1402 and three other issues. One of them is already fixed in the main. I will open PRs for all of the issues that I encountered. Below is Claude Code's summary of the first issue (#1402), and I could not find fault at the solution. If you prefer me to reduce the comments that Claude wanted to put there, let me know.

----

## The bug

`sync_objective` in `lightyear_sync/src/timeline/input.rs` computes:

```rust
let mut obj = remote + network_delay + jitter_margin - input_delay;
if obj < remote {
    obj = remote + TickDelta::from_i16(1)
}
```

The guard catches the case where `input_delay` overpowers `network_delay + jitter_margin`, but it does **not** catch the case where the raw arithmetic produces a value in the interval `(remote, remote + 1)`.

The server consumes inputs in `FixedPreUpdate`, which runs one tick *after* the tick at which the message was received in `PreUpdate`. The client must therefore run at least one tick ahead of the server to cover this processing gap.

At localhost RTT and default config (no input delay, `jitter_margin = 5 ms`, `error_margin = 1.0`):

| Tick rate | Raw `obj` | `error_margin` allows local to drift to | Floor → tick |
|---|---|---|---|
| 15 Hz | `remote + 0.075` | `remote - 0.925` | `remote − 1` |
| 100 Hz | `remote + 0.5` | `remote − 0.5` | `remote − 1` |

The client buffers an input for `local_tick` (which can be `remote − 1`). The server's `update_action_state` reads the buffer at `remote + 1` (after `FixedFirst` advances) and finds nothing.

Reported in #1402 by @nuzzles with their proposed fix.

## The fix

```diff
-    if obj < remote {
+    if obj < remote + TickDelta::from_i16(1) {
         obj = remote + TickDelta::from_i16(1)
     }
```

Floor `obj` at `remote + 1` so the worst-case drift permitted by `error_margin` still keeps local at or above the tick the server reads from.

## Test — fails before, passes after

Added a deterministic unit test to `lightyear_sync/src/timeline/input.rs::tests` that calls `sync_objective` directly against a `RemoteTimeline` at tick 100, with default config and a zero-RTT `PingManager`. The test asserts the returned value is `>= remote + 1`.

**Before this PR** (current `main`, sha 71c6be9):

```
$ cargo test --package lightyear_sync --lib sync_objective_keeps_local_at_least_one_tick_ahead_of_remote_issue_1402

test timeline::input::tests::sync_objective_keeps_local_at_least_one_tick_ahead_of_remote_issue_1402 ... FAILED

thread '...' panicked at lightyear_sync/src/timeline/input.rs:450:9:
sync_objective returned 100.07463, expected >= 101 (remote + 1). With localhost
RTT and default config, the objective must put local at least one tick ahead of
remote, otherwise the server's `update_action_state` reads from a buffer tick
the client never populated.

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 4 filtered out
```

**After this PR**:

```
$ cargo test --package lightyear_sync --lib sync_objective_keeps_local_at_least_one_tick_ahead_of_remote_issue_1402

test timeline::input::tests::sync_objective_keeps_local_at_least_one_tick_ahead_of_remote_issue_1402 ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 4 filtered out
```

## Safety / risk

**Correctness.** This is a strict tightening of the floor. The previous guard was a subset of the new one: any case where the old `obj < remote` triggered, the new `obj < remote + 1` also triggers. The previous behaviour is preserved for those cases. New behaviour applies only when `0 ≤ obj - remote < 1`, which is the bug case.

**Performance.** Zero. One additional `TickDelta::from_i16(1)` add per sync evaluation (sub-microsecond, runs at sync-tick rate). The client's local tick sits one tick further ahead of remote, which:
- adds one tick of prediction-replay distance — `O(1)` extra `FixedUpdate` per rollback worst-case;
- adds one tick of input-buffer occupancy on the server — bytes, not memory.

**Compatibility risk for other consumers.** Low. The guard is internal to `sync_objective`; downstream sees only the resulting `TickInstant`. Any project that was running correctly at the old `obj ≈ remote + epsilon` will now run at `obj = remote + 1`, which is a strictly better sync-clock position for input arrival. Projects targeting "client = server" (host-client / lockstep) take different code paths entirely and are unaffected.

**Wider test suite.** `lightyear_tests` has several pre-existing flaky tests on `main` — `test_client_use_component_history`, `test_input_broadcasting_prediction` (native), `test_compute_hash`, `test_last_confirmed_input_multiple_clients`, etc. — that I'd categorise as test-order or rate-sensitive. They pass in isolation but flake in the full suite both with and without this change. Comparing identical seeds:

| | Clean main | With this PR |
|---|---|---|
| Pass | 88 | 90 |
| Fail | 7 | 5 |
| Newly passing | — | `test_buffer_inputs_with_delay`, `test_prespawn_local_despawn_match`, `test_input_rollback_always_mode` |
| Newly failing | — | none consistently — `test_input_broadcasting_prediction` (bei) flakes, passes in isolation |

So the change improves the suite's baseline pass rate, doesn't introduce a deterministic regression, and the one new flake is pre-existing test instability that the fix unmasks rather than causes.

**Note on `test_input_delay_config` (in `lightyear_sync/src/timeline/input.rs::tests`).** This pre-existing test asserts `input_delay_ticks(rtt=300ms, tick_interval=16ms) == 12`, but on clean `main` the function returns `13`. The test fails on main without any changes. It's adjacent to the code in this PR but unrelated to the fix — flagging in case anyone wants to follow up separately.

## Related

- Server-side half of #1402 (server `update_action_state` get/pop): partially fixed in #1438 (`get` → `get_predict`); the `pop` wipe is still present. I'll open a follow-up PR for that.
- Rollback short-circuit when `confirmed_tick > tick` (separate from #1402): I'll open a separate PR with reproduction.

### DIFF
--- a/lightyear_sync/src/timeline/input.rs
+++ b/lightyear_sync/src/timeline/input.rs
@@ -310,7 +310,20 @@ impl SyncedTimeline for InputTimeline {
         // `error_margin` without correcting, so include that tolerance in the
         // objective; otherwise the controller can consume the entire delivery
         // margin in steady state.
-        let obj = remote + network_delay + jitter_margin + sync_error_margin - input_delay;
+        let mut obj = remote + network_delay + jitter_margin + sync_error_margin - input_delay;
+
+        // Floor at `remote + 1 + sync_error_margin` so even worst-case
+        // controller drift (`offset = -error_margin`) leaves the client's
+        // local tick at >= remote + 1. The server reads inputs in
+        // `FixedPreUpdate`, one tick after they arrive in `PreUpdate`, so
+        // local must be at least one tick ahead of the server at send
+        // time. Without this floor, low-latency / low-jitter setups (or
+        // any config with `jitter_margin < 1.0`) can let local drift to
+        // `remote`, causing the server to read an empty buffer entry.
+        let floor = remote + TickDelta::from_i32(1) + sync_error_margin;
+        if obj < floor {
+            obj = floor;
+        }
         trace!(
             ?remote,
             ?network_delay,
@@ -471,6 +484,64 @@ mod tests {
         let objective = timeline.sync_objective(&remote, &config, &ping_manager, tick_duration);
 
         assert_tick_instant_close(objective, TickInstant::lit("102.75"));
+    }
+
+    /// The server reads inputs in `FixedPreUpdate`, one tick after they
+    /// arrive in `PreUpdate`. The client's local tick must therefore be
+    /// at least `remote + 1` at the time the input is sent, even under
+    /// worst-case controller drift (`offset = -error_margin`).
+    ///
+    /// The post-replicon `+sync_error_margin` term in the objective
+    /// cancels with the symmetric controller drift, leaving the safety
+    /// margin riding on `network_delay + jitter_margin - input_delay`.
+    /// At localhost (zero RTT, zero jitter, zero input_delay) and a
+    /// user-tightened `jitter_margin = 0.5`, that remaining margin is
+    /// only 0.5 ticks — the worst-case local tick drifts to
+    /// `remote + 0.5`, which floors to `remote` (one tick short).
+    ///
+    /// A drift-aware floor at `remote + 1 + error_margin` keeps the
+    /// invariant explicit and robust under any `jitter_margin` config.
+    #[test]
+    fn sync_objective_keeps_local_at_least_one_tick_ahead_under_worst_case_drift() {
+        let tick_duration = Duration::from_millis(10);
+        let mut remote = RemoteTimeline::default();
+        remote.set_now(TickInstant::from(Tick(100)));
+
+        // Localhost — zero RTT, zero jitter.
+        let mut ping_manager = PingManager::default();
+        ping_manager.rtt_estimator_ewma.final_stats.rtt = Duration::ZERO;
+        ping_manager.rtt_estimator_ewma.final_stats.jitter = Duration::ZERO;
+
+        // User tightens `jitter_margin` below 1.0 for snappier sync.
+        // `error_margin` stays at the default 1.0.
+        let mut config = InputTimelineConfig::default();
+        config.sync.jitter_margin = 0.5;
+        assert!(
+            config.sync.error_margin >= 1.0,
+            "test premise: error_margin is at least 1 tick"
+        );
+
+        let objective =
+            InputTimeline::default().sync_objective(&remote, &config, &ping_manager, tick_duration);
+
+        // Controller may legitimately let `local - objective` reach
+        // `-error_margin` without correcting (see `SyncContext::speed_adjustment`).
+        let worst_case_drift = TickDelta::from_duration(
+            tick_duration.mul_f32(config.sync.error_margin),
+            tick_duration,
+        );
+        let worst_case_local = objective - worst_case_drift;
+
+        let floor = TickInstant::from(Tick(101));
+        assert!(
+            worst_case_local >= floor,
+            "worst-case local tick is {worst_case_local:?}, but the server \
+             reads inputs in FixedPreUpdate one tick after PreUpdate \
+             arrival, so local must be >= {floor:?} (= remote + 1). The \
+             current `+sync_error_margin` formula only guarantees this \
+             when `jitter_margin >= 1.0`; a drift-aware floor at \
+             `remote + 1 + error_margin` is needed.",
+        );
     }
 
     #[test]


mmannerm:
After letting my claude code review #1361, this PR’s `obj < remote` floor appears to be intentionally removed in the replicon-side rewrite of `sync_objective`:

<https://github.com/cBournhonesque/lightyear/blob/cb/lightyear-replicon/lightyear_sync/src/timeline/input.rs#L305-L313>

The replicon version computes
`obj = remote + network_delay + jitter_margin + sync_error_margin - input_delay`
with no `obj < remote` clamp, and the inline comment explains why:

> Keep the objective as-computed — including when input_delay pushes it
> behind the server — rather than nudging it to `remote + 1`. Deterministic
> replication requires the client timeline to stay close to the real
> objective so input packets for tick T arrive before the server simulates
> T. The sync controller is allowed to drift within `error_margin` without
> correcting, so include that tolerance in the objective; otherwise the
> controller can consume the entire delivery margin in steady state.

So the same problem (input arriving for a tick the server has already
processed) is addressed differently in the replicon branch — by including
`sync_error_margin` directly in the objective formula rather than flooring
the result. The `Tick` type also widens from `u16` to `u32` there.

If the replicon PR lands first, this PR becomes a conflicting design and
should be closed. Happy to do that whenever you confirm; just wanted to
flag it so we don't end up with a contradictory floor in the post-replicon
codebase. Let me know how you'd like to handle this.

cBournhonesque:
Hi, 
I've merged the replicon PR, but I think your approach here is maybe more explicit and safer?
The `sync_error_margin` approach relies on the error margin being at least 1 tick, and your PR makes it more clear that the issue comes from a one tick offset due to the reading the inputs in FixedPreUpdate.

What do you think?

Otherwise not forcing the objective to be ahead of the server might be more correct; I had this rule in place because of how lightyear's replication worked, but i'm not entirely sure if it's still necessary now with replicon.
(I ran examples successfully with input delay, so probably not?)

mmannerm:
Thanks for taking a more careful look at this, as I think you're right that we need actually both. Here is what I eventually came up after interrogating claude enough.

---

 I ran the math against the post-replicon code and I think you're right that this is the more defensible framing — but with a refinement, because **neither the current `+sync_error_margin` approach nor this PR's `obj < remote + 1` floor alone is fully sufficient**. Both protect against different cases and the safest fix combines them.

**Timing constraint still holds post-replicon.** I checked: `receive_input_message` still runs in `PreUpdate`, `update_action_state` still runs in `FixedPreUpdate`. So the client must be at least one tick ahead of the server at the time the input is sent for the input for tick T to be present when the server simulates T. The replicon refactor didn't change this ordering. Examples passing with `input_delay > 0` doesn't really test the bug — input delay provides headroom that masks it.

**The `+sync_error_margin` dependency isn't on `error_margin`.** It cancels:

```
obj = remote + network_delay + jitter_margin + sync_error_margin - input_delay
worst-case local = obj - error_margin  // controller drifts symmetrically
                 = remote + network + jitter - input_delay
```

At zero-RTT localhost / `input_delay = 0`: worst-case local = `remote + jitter_margin`. The default `jitter_margin = 1.0` makes this `remote + 1.0` exactly — the bug is just-barely covered, and any user tightening `jitter_margin < 1.0` re-opens it. So the current code's safety rides on a default-config value.

**This PR's floor alone is also insufficient.** With `if obj < remote + 1 { obj = remote + 1 }`, worst-case local is `obj - error_margin = remote + 1 - 1.0 = remote`. The floor protects `obj` but not the controller-drift floor.

**Suggested fix — combine them, with a drift-aware floor:**

```rust
let obj = remote + network_delay + jitter_margin + sync_error_margin - input_delay;
// Floor at remote + 1 + error_margin so even worst-case controller
// drift leaves local at >= remote + 1 (the tick the server reads in
// FixedPreUpdate).
let floor = remote_tick + TickDelta::from_i16(1) + sync_error_margin;
if obj < floor { obj = floor }
```

This is robust under any `jitter_margin` / `error_margin` combination and names the FixedPreUpdate timing explicitly. I'll also tighten the test to assert `objective - error_margin >= remote + 1` (the controller-drift-aware version) rather than just `objective >= remote + 1`.

I will rebase this fix, and try to verify with tests.

mmannerm:
Pushed the rebase + drift-aware floor.

----

A couple of things worth noting from the implementation:

- **TDD-first verified.** I wrote the failing test against bare `upstream/main` first to confirm the bug is reachable on current code, not just on the pre-replicon version. With `jitter_margin = 0.5` (a representative tightened-config), worst-case local lands at `100.5` on bare main — below the `>= 101` invariant. After the floor, it lands at `101.5`. Failure message names the FixedPreUpdate / PreUpdate offset directly so it's easy to interpret if it ever fires again.

- **Why `jitter_margin = 0.5` for the test, not the default 1.0.** At default `jitter_margin = 1.0` with zero RTT / zero jitter, the math gives worst-case local = exactly `remote + 1.0` — the bug is just-barely covered at the boundary. A test at default config would only catch a future regression that *also* moves the boundary; `jitter_margin = 0.5` exercises the "user tightens config for snappier sync" path that already reaches the bug today.

- **No regressions on the existing sync tests.** `input_timeline_objective_preserves_margin_after_sync_deadband` and `input_delay_still_offsets_input_timeline_objective` both continue to assert their original expected `obj` values unchanged. The floor only fires when `raw_obj < remote + 1 + sync_error_margin`, which is below the regime those tests cover (moderate latency + the default jitter/error config). So this is a pure tightening — no behaviour change in the working-fine cases.

- **Full `lightyear_tests` suite: 122 passed, 0 failed (serial).** No regressions outside `lightyear_sync` either.

Happy to iterate further if you'd like the floor expressed differently (e.g. dropping `+sync_error_margin` from `obj` and rolling it into the floor as `remote + 1 + sync_error_margin` only — same worst-case behaviour, slightly different intermediate shape).

mmannerm:
Should be ready for merge now

cBournhonesque:
Thanks!
I'm not convinced that `let floor = remote + TickDelta::from_i32(1) + sync_error_margin;` was correct, it doesn't take into account input delay. I instead just added `1` in the main objective timestamp, and updated the comments. 
Will close this and merge https://github.com/cBournhonesque/lightyear/pull/1479

PR #1473, opened 2026-05-13
Restrict prepare-input rate

### LINKED ISSUE #1469: InputConfig::send_interval not used
`InptuConfig::send_interval` is not used. I expect it to set how often lightyear sends input over the network. Currently, inputs are sent over the network as fast as possible, regardless of `send_interval`'s value, which uses a huge amount of bandwidth.

Here's the setup:
1. I add the `lightyear::lightyear_inputs_bei::InputPlugin` for my component:
```rust
#[derive(Component, Serialize, Deserialize, Clone, Copy, Debug, PartialEq, Reflect)]
pub struct Foo;

app.add_plugins(InputPlugin::<Foo> {
    config: InputConfig::<Foo> {
        send_interval: Duration::from_millis(16),
        ..default()
    },
});
```

2. `lightyear::lightyear_inputs_bei::InputPlugin::build()` adds a `ClientInputPlugin`
3. `ClientInputPlugin::build()` adds the `lightyear::lightyear_inputs::InputPlugin` (does not pass the input config)
3.a. It also inserts `InputConfig` as a resource and its `send_interval` is never used.
4. `lightyear::lightyear_inputs::InputPlugin::build()` adds an `InputChannel` channel
4.a. `InputChannel::send_frequency` is set to 0 and **not** `InptuConfig::send_interval`.

nick-e:
Restrict the `prepare_input_message` system to run every `InputConfig::send_interval`. Currently, it runs every frame resulting in input data being sent to peers every frame. This change restricts the input send rate to `send_interval` and allows users to reduce bandwidth usage.

Fixes #1469


cBournhonesque:
Thanks! 
Good insight that we channel itself has to send messages as soon as possible after they were buffered to minimize input latency.

Issue #1347, opened 2025-12-05
Add visibility unit test


cBournhonesque:
With the ReplicationState refactor, there is a bug if in `lose_visibility` we have
`let state = self.per_sender_state.entry(sender).or_default();`
but it works with 
`let state = self.per_sender_state.entry(sender).or_insert_with(PerSenderReplicationState::with_authority);`

It's because if we don't have authority, then we cannot send a despawn message.

What happens is:
- we add Replicate for entity 1 -> we have explicit authority
- we lose_authority for entity 1, then update() -> ReplicationState is cleared
- we call lose_authority again for entity 1 -> we add ReplicationState with no authority
- we call gain_authority for entity 1 -> we update ReplicationState, but without authority!

The real issue is that `lose_authority` clears the PerSenderReplicationState for this entity, but that could contain relevant information about authority, etc.

Potential solution
- not remove the PerSenderReplicationState after we lose visibility. The problem is that we could fill the map with extra information that is not relevant. We have the same issue with authority: if we receive an entity from Replication and mark it as `no-authority` we add it to ReplicationState (which clutters the map) even though it's not. We could split this information into multiple components but that brings us back to where we are now...
- maybe we can switch to assigning an integer for each connected link, and we can store cheap BitSets for each data (VisibilityState, Authority, etc.) ?
- for now, just give authority to after lose_vis/give_vis if there was no data in ReplicationState (but that is NOT correct if we had no authority over this entity)

jhash:
Hi @cBournhonesque! I am new to the project and wanted to tackle some low-hanging fruit. I added a unit test for the changes you made to fix the gain_authority call with an empty `per_sender_state` to begin with. Let me know what you think.

https://github.com/cBournhonesque/lightyear/pull/1484

PR #1484, opened 2026-05-23
Add gain authority regression unit test for issue #1347

jhash:
Addresses the second part of this issue with a unit test for the gain authority method: https://github.com/cBournhonesque/lightyear/issues/1347

With [`.or_insert_with(PerSenderReplicationState::with_authority);` in `#gain_authority`](https://github.com/cBournhonesque/lightyear/pull/1361/changes#diff-5b9818d41607b05f281fe8b75c078301de678e2caed7cdc201da992a690b4278R156) commented out: `assertion failed: state.has_authority(sender)`

With [that change](https://github.com/cBournhonesque/lightyear/pull/1361/changes#diff-5b9818d41607b05f281fe8b75c078301de678e2caed7cdc201da992a690b4278R156):
`test send::tests::gain_authority_without_existing_entry ... ok`

### DIFF
--- a/lightyear_replication/src/send.rs
+++ b/lightyear_replication/src/send.rs
@@ -834,6 +834,15 @@ mod tests {
         assert_eq!(app.world().resource::<ServerTick>().get(), 2);
         assert_eq!(app.world().resource::<SendCount>().0, 2);
     }
+
+    #[test]
+    fn gain_authority_without_existing_entry() {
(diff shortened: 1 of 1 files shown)


cBournhonesque:
Thanks, that works for me!
Although with the recent move to bevy_replicon, the authority features become unused/unsupported (because bevy_replicon doesn't support client->server replication; that will come in the future versions)

PR #1488, opened 2026-05-28
Add replicate-once component registration

monsterrdev:
## Summary

Adds Lightyear component-registration wrappers for Replicon's `ReplicationMode::Once`.

This restores the Lightyear-facing API for components that should replicate insertion/removal lifecycle events without sending ordinary mutations, while keeping the implementation aligned with the Replicon integration described in `lightyear_replication/src/REPLICON_INTEGRATION.md`.

## Changes

- Add `AppComponentExt::register_component_once::<C>()`
- Add `AppComponentExt::register_component_once_with::<C>(...)` for custom Replicon serde
- Share Lightyear component metadata registration across default, once, custom, and custom-once registration paths
- Update deterministic replication protocol setup to use the Lightyear wrapper instead of calling Replicon directly
- Remove stale Replicon integration TODO notes that said replicate-once was not wired
- Add coverage for replicate-once insertion, mutation suppression, removal, and custom-serde registration

## Notes

`ReplicationMode::Once` is implemented through Replicon's rule API rather than resurrecting the old Lightyear `ComponentReplicationConfig` path. This keeps the public Lightyear API small while following the direction of the Replicon migration.

The custom-serde test component intentionally does not derive `Serialize` or `Deserialize`, so the test verifies that quantized/custom component representations can still be registered through the new wrapper.

## Testing

- `cargo fmt --all`
- `cargo check -p lightyear_replication`
- `cargo check -p deterministic_replication`
- `cargo test -p lightyear_tests test_component_replicate_once -- --nocapture`
- `git diff --check`

### DIFF
--- a/examples/deterministic_replication/src/protocol.rs
+++ b/examples/deterministic_replication/src/protocol.rs
@@ -1,6 +1,5 @@
 use avian2d::prelude::*;
 use bevy::prelude::*;
-use bevy_replicon::prelude::AppRuleExt;
 use leafwing_input_manager::prelude::*;
 use lightyear::input::config::InputConfig;
 use lightyear::prelude::input::leafwing;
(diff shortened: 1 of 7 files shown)


cBournhonesque:
Thanks this looks great!

PR #1505, opened 2026-06-10
Keep prediction history sorted when predicted samples follow future confirmations

monsterrdev:
## Summary

Fix `PredictionHistory` ordering when a predicted sample is recorded after a confirmed sample from a future tick.

With the Replicon-backed prediction path, confirmed component updates can be written into `PredictionHistory` ahead of the client's current local prediction tick. While the client catches up, local prediction can then record older predicted ticks.

Previously, predicted samples were always appended, which could produce an unsorted history like:

```text
670 predicted
691 confirmed
676 predicted
```

`PredictionHistory` already relies on sorted tick order for `partition_point()` lookups, rollback restore, correction, and pruning. Once the buffer is unsorted, those reads can return the wrong state.

## Changes

- Keep the normal monotonic predicted path as an O(1) append.
- Use sorted insertion when a predicted sample arrives older than the current newest history entry.
- Make `second_most_recent(tick)` tick-relative so future confirmed entries are ignored when computing the previous visual-correction value.
- Add a regression test covering predicted insertion after a future confirmed sample.

## Notes

This is a narrow invariant fix for `PredictionHistory`. It does not attempt to address the broader Replicon rollback-detection follow-ups listed in `REPLICON_INTEGRATION.md`.

## Testing

```text
cargo fmt -- --check
cargo test -p lightyear_prediction predicted_history::tests
cargo test -p lightyear_prediction --lib
```

### DIFF
--- a/lightyear_prediction/src/predicted_history.rs
+++ b/lightyear_prediction/src/predicted_history.rs
@@ -285,30 +285,44 @@ impl<C> PredictionHistory<C> {
     ///
     /// Used to efficiently get the previous value when doing correction.
     pub fn second_most_recent(&self, tick: Tick) -> Option<&C> {
-        let (most_recent_tick, most_recent) = self.most_recent()?;
-        if *most_recent_tick < tick {
-            return most_recent.into();
+        if let Some((most_recent_tick, most_recent)) = self.most_recent()
+            && *most_recent_tick < tick
+        {
+            return most_recent.value();
         }
-        let len = self.buffer.len();
-        if len < 2 {
+        let partition = self
+            .buffer
+            .partition_point(|(buffer_tick, _)| *buffer_tick <= tick);
+        if partition == 0 {
             return None;
         }
-        self.buffer.get(len - 2).and_then(|(_, x)| x.into())
+        let latest_at_or_before_tick = partition - 1;
+        if self.buffer[latest_at_or_before_tick].0 < tick {
+            return self.buffer[latest_at_or_before_tick].1.value();
+        }
+        if latest_at_or_before_tick == 0 {
+            return None;
+        }
+        self.buffer
+            .get(latest_at_or_before_tick - 1)
+            .and_then(|(_, state)| state.value())
     }
 
-    /// Add a value with a specific state (for predicted values, appends to end)
+    /// Add a value with a specific state.
     fn add_state(&mut self, tick: Tick, state: PredictionState<C>) {
-        if let Some((last_tick, _)) = self.buffer.back()
-            && *last_tick == tick
-        {
-            // Replace the existing value at this tick
-            self.buffer.pop_back();
+        match self.buffer.back().map(|(last_tick, _)| *last_tick) {
+            None => self.buffer.push_back((tick, state)),
+            Some(last_tick) if last_tick < tick => self.buffer.push_back((tick, state)),
+            Some(last_tick) if last_tick == tick => {
+                self.buffer.pop_back();
+                self.buffer.push_back((tick, state));
+            }
+            Some(_) => self.insert_at_tick(tick, state),
         }
-        self.buffer.push_back((tick, state));
     }
 
     /// Insert a value at the correct position in the buffer (maintaining tick order).
-    /// This is used for confirmed values which might arrive out of order.
+    /// This is used for values that might arrive out of order.
     /// If a value already exists at this tick, it will be replaced.
     fn insert_at_tick(&mut self, tick: Tick, state: PredictionState<C>) {
         // Find the position where this tick should be inserted
@@ -753,5 +767,28 @@ mod tests {
                 .0,
             100.0
         );
+    }
+
+    #[test]
+    fn predicted_insert_after_future_confirmed_preserves_tick_order() {
+        let mut history = PredictionHistory::<TestValue>::default();
+
+        history.add_predicted(Tick(670), Some(TestValue(670.0)));
+        history.add_confirmed(Tick(691), Some(TestValue(691.0)));
+        history.add_predicted(Tick(676), Some(TestValue(676.0)));
+
+        assert_eq!(history.buffer().len(), 3);
+        assert_eq!(history.buffer()[0].0, Tick(670));
+        assert_eq!(history.buffer()[1].0, Tick(676));
+        assert_eq!(history.buffer()[2].0, Tick(691));
+        assert_eq!(
+            history.most_recent().map(|(tick, _)| *tick),
+            Some(Tick(691))
+        );
+        assert_eq!(history.get(Tick(680)).unwrap().0, 676.0);
+        assert_eq!(history.get(Tick(691)).unwrap().0, 691.0);
+        assert_eq!(history.second_most_recent(Tick(676)).unwrap().0, 670.0);
+        assert_eq!(history.second_most_recent(Tick(680)).unwrap().0, 676.0);
+        assert_eq!(history.second_most_recent(Tick(691)).unwrap().0, 676.0);
     }
 }
--- a/lightyear_prediction/src/predicted_history.rs
+++ b/lightyear_prediction/src/predicted_history.rs
@@ -8,6 +8,7 @@
 use crate::rollback::DeterministicPredicted;
 use crate::{Predicted, SyncComponent};
 use alloc::collections::VecDeque;
+use alloc::vec::Vec;
 use bevy_ecs::component::Mutable;
 use bevy_ecs::prelude::*;
 use bevy_reflect::Reflect;
@@ -123,23 +124,27 @@ impl<R> From<PredictionState<R>> for Option<R> {
 /// 3. Avoid re-predicting values we already know are correct from the server
 #[derive(Component, Debug, Reflect)]
 pub struct PredictionHistory<C> {
-    /// Queue containing the history. Front = oldest, back = most recent.
-    /// Only stores ticks where the value changed.
-    buffer: VecDeque<(Tick, PredictionState<C>)>,
+    /// Locally computed values. Front = oldest, back = most recent.
+    /// These are cleared and rebuilt during rollback.
+    predicted: VecDeque<(Tick, PredictionState<C>)>,
+    /// Server-authoritative values. Front = oldest, back = most recent.
+    /// These are preserved during rollback and can arrive out of order.
+    confirmed: VecDeque<(Tick, PredictionState<C>)>,
 }
 
 impl<C> Default for PredictionHistory<C> {
     fn default() -> Self {
         Self {
-            buffer: VecDeque::new(),
+            predicted: VecDeque::new(),
+            confirmed: VecDeque::new(),
         }
     }
 }
 
-impl<C: Debug> Display for PredictionHistory<C> {
+impl<C: Debug + Clone> Display for PredictionHistory<C> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "PredictionHistory[")?;
-        for (i, (tick, state)) in self.buffer.iter().enumerate() {
+        for (i, (tick, state)) in self.merged_entries().iter().enumerate() {
             if i > 0 {
                 write!(f, ", ")?;
             }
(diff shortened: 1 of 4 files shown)


cBournhonesque:
Is this something you have observed in a real example?
Would it be possible to have an integration test for it?

It should be very rare/impossible to receive a confirmed update in the future compared to the predicted updates.

monsterrdev:
>Is this something you have observed in a real example?

I have a small bevy + lightyear project that I am intentionally stress testing to see which codepaths scale poorly.

The stress test is 100 entities (NPCs moving around) all replicating things like position, velocity, facing, etc... Some of the tick systems are causing intermittent server tick overrun at a lower interval (4hz). There's only one client connected for this scenario. I haven't looked much into replicon (I assume it should be the same), but under lightyear this produced a lot of chunked/split messages (not fragmented) as the amount of data I send when the server is being stressed is more than what fits into a single 1200 packet per tick.

I'm running the client and server on the same host machine without any link conditioning applied.

The setup is a 64 Hz sim with 32 Hz replication sends. Under load/catch-up, the client ended up with confirmed history for a newer tick, then normal prediction/replay recorded an older predicted tick afterward. In one captured case the shape was effectively:

- predicted history around tick 670
- confirmed update at tick 691
- later predicted write at tick 676

The issue was not that the server was sending impossible future data in wall-clock terms, but that `PredictionHistory` could receive a confirmed sample that was future-relative to a later local predicted history insert. Because predicted inserts appended, the buffer became unsorted, and `partition_point`/`second_most_recent` lookups could read the wrong state.

>Would it be possible to have an integration test for it?

Yup, abc9f2f42605818fa952dfc01ff464d68ea2ddc0 let me know if you had something different in mind.

>It should be very rare/impossible to receive a confirmed update in the future compared to the predicted updates.

I agree this should be rare in steady state, but rollback/replay/catch-up makes it reachable.

### Extra context

The original symptom I was debugging was remote NPCs visually freezing/desyncing while the server kept advancing them. I added temporary diagnostics for replicated NPC positions, confirmed/interpolated history ticks, packet sizes, split replication messages, and client-side history skips. This allowed me to compare what the server was sending against what the client was receiving. Those pointed to a Lightyear 0.18 issue where split update messages from the same tick were being received, but later chunks were treated as non-history updates, causing many component updates to be skipped.

I had a [local fix](https://github.com/cBournhonesque/lightyear/compare/main...monsterrdev:lightyear:fix/split-updates-same-tick) for that path, but remembered main had moved to the Replicon integration, so I tested current main instead. Replicon fixed the original split-message/NPC-stall issue I was seeing. but I noticed player movement was sluggish in comparison to what it was in 0.18. Using the same diagnostics, I then found the prediction-history issue described in this PR: during correction/replay, a newer confirmed sample could already be present, and normal prediction could later record an older predicted sample behind it. That left the history buffer out of tick order and broke tick-relative lookups used by correction.

So the PR issue was found on current main, but it came from a real gameplay reproduction rather than a synthetic case.

cBournhonesque:
I see; I still don't fully understand how it is possible that a Confirmed server update is received for a tick that is a future compared to the latest predicted tick, but in any case it is good to make this path more robust.

I will review the PR more carefully later this week.
Thanks!

lightyear_prediction/src/predicted_history.rs
```diff
@@ -285,30 +285,44 @@ impl<C> PredictionHistory<C> {
     ///
     /// Used to efficiently get the previous value when doing correction.
     pub fn second_most_recent(&self, tick: Tick) -> Option<&C> {
-        let (most_recent_tick, most_recent) = self.most_recent()?;
-        if *most_recent_tick < tick {
-            return most_recent.into();
+        if let Some((most_recent_tick, most_recent)) = self.most_recent()
```
cBournhonesque: I don't remember why this logic was necessary, but it doesn't seem correct?
I guess it might be because we might be running correction without having added a new value for the current tick in the history, in which case the second-most-recent value is the last value in the buffer?

That seems a bit flimsy

lightyear_prediction/src/predicted_history.rs
```diff
@@ -285,30 +285,44 @@ impl<C> PredictionHistory<C> {
     ///
     /// Used to efficiently get the previous value when doing correction.
     pub fn second_most_recent(&self, tick: Tick) -> Option<&C> {
```
cBournhonesque, line 334: Maybe this should be renamed `second_most_recent_before_tick`, if we are ignoring confirmed-values that have been added for ticks after `tick`

cBournhonesque:
Could you please share your stress-test? 
I didn't understand from your explanation why we would receive a confirmed tick from the future. 
I guess one reason it could happen is if the timeline syncs break down

lightyear_prediction/src/predicted_history.rs
```diff
@@ -285,30 +285,44 @@ impl<C> PredictionHistory<C> {
...
-        if let Some((last_tick, _)) = self.buffer.back()
-            && *last_tick == tick
-        {
-            // Replace the existing value at this tick
-            self.buffer.pop_back();
+        match self.buffer.back().map(|(last_tick, _)| *last_tick) {
+            None => self.buffer.push_back((tick, state)),
+            Some(last_tick) if last_tick < tick => self.buffer.push_back((tick, state)),
+            Some(last_tick) if last_tick == tick => {
+                self.buffer.pop_back();
```
cBournhonesque: I know this was the previous behaviour, but this seems incorrect, no?
We never want a Predicted value to replace a Confirmed value; if we had an existing Confirmed value at tick T, we should be keeping that.

Althought we might want to actually store both the Predicted and Confirmed value?
We need the Confirmed for sure for future rollback checks, to restore to the correct value at that tick.
We might want the Predicted value so that we can issue a rollback correctly?

Maybe it's a design issue and we should have a ConfirmedHistory separately from the PredictedHistory

monsterrdev:
I've found the reproduction has been sensitive to CPU load, so if you're running it with a beefy computer that is largely idle I wouldn't be surprised if you didn't see the issue pop up. I'm building this on a little M1, and the actual project has physics, npc pathing, rendering, and everything else that goes with it, so I was seeing it steady state.

## Expected Output

Leave the client idle, and you'll see. 

```text
2026-06-11T19:27:09.204655Z  WARN client: PredictionHistory<Position> out of order: count=1 local_tick=Tick(20) pair=(Tick(66), confirmed) -> (Tick(20), predicted)
history=66C, 20P

2026-06-11T19:27:10.216758Z  WARN client: PredictionHistory<Position> out of order: count=4279 local_tick=Tick(127) pair=(Tick(126), confirmed) -> (Tick(90), confirmed)
history=124C, 126C, 90C, 92C, 128C, 130C, 133C, 134C, 136C, 138C, 93P, 90P, 91P, 92P, 93P, 94P, 95C, 96C, 97P, 98C, 99P, 100C, 101P, 102C, 103P, 104C, 105P, 106P, 107C, 108C, 109P, 110C, 111P, 112C, 113P, 114C, 115P, 116C, 117P, 118P, 119C, 120C, 121P, 122C, 123P, 124C, 125P, 126P, 127P
```

In the compact history string:

- `C` means confirmed
- `P` means predicted
- `R` means removed/other

The important part is a later tick followed by an earlier tick, for example:

```text
history=54276C, 54275P, 54276P, 54277P
```

That means the history buffer is no longer sorted.

## Directory Layout

```text
repro_lightyear_prediction_history_ordering/
  Cargo.toml
  src/
    lib.rs
    bin/
      server.rs
      client.rs
```

## Running

```sh
cargo run --bin server --release
```

Terminal 2:

```sh
cargo run --bin client --release
```

## Cargo.toml

```toml
[package]
name = "repro_lightyear_prediction_history_ordering"
version = "0.1.0"
edition = "2024"

# Standalone repro. It is intentionally not a member of any parent workspace so
# it can be copied into a Lightyear fork or attached to an upstream issue.
[workspace]

[dependencies]
bevy = { version = "0.18", default-features = false, features = [
    "async_executor",
    "bevy_log",
    "bevy_state",
    "multi_threaded",
    "std",
] }
bevy_replicon = "0.39"
lightyear = { version = "0.26.4", features = ["udp", "netcode", "input_native", "frame_interpolation"] }
serde = { version = "1", features = ["derive"] }
glam = { version = "0.30", features = ["serde"] }

[[bin]]
name = "server"
path = "src/bin/server.rs"

[[bin]]
name = "client"
path = "src/bin/client.rs"

[lib]
path = "src/lib.rs"

# Point at the latest upstream-main Lightyear checkout used for the replicon
# comparison.
[patch.crates-io]
lightyear = { path = "your_path_to_lightyear/lightyear" }
```

## src/lib.rs

```rust
use bevy::prelude::*;
use lightyear::prelude::*;
use serde::{Deserialize, Serialize};

pub const SERVER_PORT: u16 = 5020;
pub const PROTOCOL_ID: u64 = 2;
pub const PRIVATE_KEY: [u8; 32] = [1u8; 32];

pub const SERVER_TICK_HZ: f64 = 64.0;
pub const INTERPOLATED_ENTITY_COUNT: usize = 100;

#[derive(Component, Serialize, Deserialize, Clone, Copy, Debug, Default, PartialEq)]
pub struct Position(pub Vec3);

#[derive(Component, Serialize, Deserialize, Clone, Copy, Debug, Default, PartialEq)]
pub struct Velocity(pub Vec3);

#[derive(Component, Serialize, Deserialize, Clone, Copy, Debug, Default, PartialEq)]
pub struct Facing {
    pub yaw: f32,
    pub yaw_rate: f32,
    pub pitch: f32,
    pub flags: u8,
}

pub struct ProtocolPlugin;

impl Plugin for ProtocolPlugin {
    fn build(&self, app: &mut App) {
        app.register_component::<Position>()
            .add_prediction()
            .add_interpolation_with(position_lerp);
        app.register_component::<Velocity>()
            .add_prediction()
            .add_interpolation_with(velocity_lerp);
        app.register_component::<Facing>()
            .add_prediction()
            .add_interpolation_with(facing_lerp);
    }
}

fn position_lerp(start: Position, end: Position, t: f32) -> Position {
    Position(start.0.lerp(end.0, t))
}

fn velocity_lerp(start: Velocity, end: Velocity, t: f32) -> Velocity {
    Velocity(start.0.lerp(end.0, t))
}

fn facing_lerp(start: Facing, end: Facing, t: f32) -> Facing {
    Facing {
        yaw: start.yaw.lerp(end.yaw, t),
        yaw_rate: start.yaw_rate.lerp(end.yaw_rate, t),
        pitch: start.pitch.lerp(end.pitch, t),
        flags: if t < 0.5 { start.flags } else { end.flags },
    }
}

pub fn tick_duration() -> std::time::Duration {
    std::time::Duration::from_secs_f64(1.0 / SERVER_TICK_HZ)
}
```

## src/bin/server.rs

```rust
use std::net::{Ipv4Addr, SocketAddr};
use std::time::Duration;

use bevy::app::ScheduleRunnerPlugin;
use bevy::log::LogPlugin;
use bevy::prelude::*;
use bevy::state::app::StatesPlugin;
use lightyear::prelude::server::*;
use lightyear::prelude::*;

use repro_lightyear_prediction_history_ordering::{
    Facing, INTERPOLATED_ENTITY_COUNT, PRIVATE_KEY, PROTOCOL_ID, Position, ProtocolPlugin,
    SERVER_PORT, SERVER_TICK_HZ, Velocity, tick_duration,
};

const REPLICATION_HZ: f64 = 32.0;

fn replication_interval() -> Duration {
    Duration::from_secs_f64(1.0 / REPLICATION_HZ)
}

#[derive(Component)]
struct MotionEntity {
    index: usize,
}

#[derive(Component)]
struct InterpolatedLoadEntity;

#[derive(Component)]
struct PredictedPlayer;

#[derive(Resource, Default)]
struct SimTick(u64);

fn main() {
    App::new()
        .add_plugins(MinimalPlugins.set(ScheduleRunnerPlugin::run_loop(tick_duration())))
        .add_plugins(StatesPlugin)
        .add_plugins(LogPlugin::default())
        .add_plugins(ServerPlugins {
            tick_duration: tick_duration(),
        })
        .add_plugins(ProtocolPlugin)
        .insert_resource(ReplicationMetadata::new(replication_interval()))
        .init_resource::<SimTick>()
        .add_systems(Startup, start_server)
        .add_observer(setup_client_replication)
        .add_observer(spawn_predicted_player)
        .add_systems(
            FixedUpdate,
            (update_authoritative_entities).chain(),
        )
        .run();
}

fn start_server(mut commands: Commands) {
    let server_addr = SocketAddr::new(Ipv4Addr::UNSPECIFIED.into(), SERVER_PORT);
    let server_entity = commands
        .spawn((
            Name::new("Server"),
            NetcodeServer::new(
                NetcodeConfig::default()
                    .with_protocol_id(PROTOCOL_ID)
                    .with_key(PRIVATE_KEY),
            ),
            ServerUdpIo::default(),
            LocalAddr(server_addr),
        ))
        .id();

    spawn_dense_npc_pack(&mut commands);

    info!(
        "server: listening on {server_addr}; tick={SERVER_TICK_HZ}Hz replication={REPLICATION_HZ}Hz interpolated_entities={INTERPOLATED_ENTITY_COUNT}"
    );
    commands.trigger(Start {
        entity: server_entity,
    });
}

fn spawn_dense_npc_pack(commands: &mut Commands) {
    for index in 0..INTERPOLATED_ENTITY_COUNT {
        let x = (index % 10) as f32;
        let z = (index / 10) as f32;
        commands.spawn((
            Name::new(format!("InterpolatedLoadEntity{index}")),
            InterpolatedLoadEntity,
            MotionEntity { index },
            Position(Vec3::new(x, 0.0, z)),
            Velocity(Vec3::ZERO),
            Facing::default(),
            Replicate::to_clients(NetworkTarget::All),
            InterpolationTarget::to_clients(NetworkTarget::All),
        ));
    }
}

fn setup_client_replication(trigger: On<Add, LinkOf>, mut commands: Commands) {
    commands.entity(trigger.entity).insert(ReplicationSender);
    info!(
        "server: client linked; replication sender interval {:?}",
        replication_interval()
    );
}

fn spawn_predicted_player(
    trigger: On<Add, Connected>,
    client_links: Query<&RemoteId, With<ClientOf>>,
    mut commands: Commands,
) {
    let Ok(remote_id) = client_links.get(trigger.entity) else {
        return;
    };
    let client_id = remote_id.0;
    let index = INTERPOLATED_ENTITY_COUNT;

    commands.spawn((
        Name::new("PredictedPlayer"),
        PredictedPlayer,
        MotionEntity { index },
        Position(Vec3::new(0.0, 0.0, -4.0)),
        Velocity(Vec3::ZERO),
        Facing::default(),
        Replicate::to_clients(NetworkTarget::All),
        PredictionTarget::to_clients(NetworkTarget::Single(client_id)),
        InterpolationTarget::to_clients(NetworkTarget::AllExceptSingle(client_id)),
    ));

    info!("server: spawned one predicted player for client {client_id:?}");
}

fn update_authoritative_entities(
    mut sim_tick: ResMut<SimTick>,
    mut query: Query<(&MotionEntity, &mut Position, &mut Velocity, &mut Facing)>,
) {
    sim_tick.0 += 1;
    let tick = sim_tick.0 as f32;

    for (entity, mut position, mut velocity, mut facing) in query.iter_mut() {
        let phase = tick * 0.04 + entity.index as f32 * 0.17;
        let speed = 0.025 + (entity.index % 7) as f32 * 0.0025;
        let next_velocity = Vec3::new(phase.cos() * speed, 0.0, phase.sin() * speed);

        position.0 += next_velocity;
        velocity.0 = next_velocity;
        facing.yaw = next_velocity.z.atan2(next_velocity.x);
        facing.yaw_rate = phase.cos() * 0.5;
        facing.pitch = (phase * 0.25).sin() * 0.05;
        facing.flags = (entity.index % 4) as u8;
    }
}
```

## src/bin/client.rs

```rust
use bevy::prelude::*;
use bevy_replicon::prelude::ClientSystems;
use lightyear::interpolation::timeline::InterpolationConfig;
use lightyear::prelude::client::*;
use lightyear::prelude::*;
use std::net::{Ipv4Addr, SocketAddr};

use repro_lightyear_prediction_history_ordering::{
    Facing, PRIVATE_KEY, PROTOCOL_ID, Position, ProtocolPlugin,
    SERVER_PORT, Velocity, tick_duration,
};

#[derive(Resource, Default, Debug)]
struct ClientProbe {
    unordered_observations: u64,
    last_reported_unordered_observations: u64,
    latest_unordered: Option<UnorderedHistoryObservation>,
    next_summary_at_secs: f64,
}

#[derive(Debug)]
struct UnorderedHistoryObservation {
    local_tick: Tick,
    before_tick: Tick,
    before_kind: &'static str,
    after_tick: Tick,
    after_kind: &'static str,
    compact_history: String,
}

fn main() {
    let client_id: u64 = 42;
    let server_addr = SocketAddr::new(Ipv4Addr::LOCALHOST.into(), SERVER_PORT);
    let client_addr = SocketAddr::new(Ipv4Addr::UNSPECIFIED.into(), 0);

    App::new()
        .add_plugins(DefaultPlugins)
        .add_plugins(ClientPlugins {
            tick_duration: tick_duration(),
        })
        .add_plugins(ProtocolPlugin)
        .init_resource::<ClientProbe>()
        .add_systems(Startup, move |mut commands: Commands| {
            let client = NetcodeClient::new(
                Authentication::Manual {
                    server_addr,
                    client_id,
                    private_key: PRIVATE_KEY,
                    protocol_id: PROTOCOL_ID,
                },
                NetcodeConfig::default(),
            )
            .expect("failed to create netcode client");

            let entity = commands
                .spawn((
                    Name::new("Client"),
                    client,
                    UdpIo::default(),
                    LocalAddr(client_addr),
                    PredictionManager::default(),
                    ReplicationReceiver::default(),
                    InterpolationConfig::default().with_send_interval_ratio(2.0),
                    InputTimelineConfig::new(SyncConfig::default(), InputDelayConfig::balanced()),
                ))
                .id();

            info!("client: connecting to {server_addr}");
            commands.trigger(Connect { entity });
        })
        .add_systems(Update, log_probe_summary)
        .add_systems(FixedUpdate, (predict_components_locally).chain())
        .add_systems(
            PreUpdate,
            inspect_position_history
                .after(ClientSystems::Receive)
                .before(RollbackSystems::Check),
        )
        .add_systems(
            FixedPostUpdate,
            inspect_position_history.after(PredictionSystems::UpdateHistory),
        )
        .run();
}

fn log_probe_summary(time: Res<Time>, mut probe: ResMut<ClientProbe>) {
    if probe.unordered_observations == 0 {
        return;
    }
    if probe.unordered_observations == probe.last_reported_unordered_observations {
        return;
    }

    let now = time.elapsed_secs_f64();
    if now < probe.next_summary_at_secs {
        return;
    }
    probe.next_summary_at_secs = now + 1.0;
    probe.last_reported_unordered_observations = probe.unordered_observations;

    if let Some(latest) = probe.latest_unordered.as_ref() {
        warn!(
            "PredictionHistory<Position> out of order: count={} local_tick={:?} pair=({:?}, {}) -> ({:?}, {}) \nhistory={}",
            probe.unordered_observations,
            latest.local_tick,
            latest.before_tick,
            latest.before_kind,
            latest.after_tick,
            latest.after_kind,
            latest.compact_history,
        );
    }
}

fn predict_components_locally(
    timeline: Res<LocalTimeline>,
    mut query: Query<(Entity, &mut Position, &mut Velocity, &mut Facing), With<Predicted>>,
) {
    let tick = timeline.tick().0 as f32;
    for (entity, mut position, mut velocity, mut facing) in query.iter_mut() {
        let seed = entity.to_bits() as f32 * 0.013;
        let phase = tick * 0.06 + seed;
        let local_velocity = Vec3::new(phase.sin() * 0.01, 0.0, phase.cos() * 0.01);

        position.0 += local_velocity;
        velocity.0 = local_velocity;
        facing.yaw += 0.01;
        facing.yaw_rate = 0.01;
    }
}

fn inspect_position_history(
    timeline: Res<LocalTimeline>,
    mut probe: ResMut<ClientProbe>,
    query: Query<(Entity, &PredictionHistory<Position>), With<Predicted>>,
) {
    let local_tick = timeline.tick();

    for (_, history) in query.iter() {
        let entries = history
            .buffer()
            .iter()
            .map(|(tick, state)| (*tick, state.clone()))
            .collect::<Vec<_>>();
        if let Some(pair) = entries.windows(2).find(|pair| pair[0].0 > pair[1].0) {
            let before_kind = if pair[0].1.is_confirmed() {
                "confirmed"
            } else if pair[0].1.is_predicted() {
                "predicted"
            } else {
                "removed"
            };
            let after_kind = if pair[1].1.is_confirmed() {
                "confirmed"
            } else if pair[1].1.is_predicted() {
                "predicted"
            } else {
                "removed"
            };
            let compact_history = entries
                .iter()
                .map(|(tick, state)| {
                    let kind = if state.is_confirmed() {
                        "C"
                    } else if state.is_predicted() {
                        "P"
                    } else {
                        "R"
                    };
                    format!("{}{}", tick.0, kind)
                })
                .collect::<Vec<_>>()
                .join(", ");
            let observation = UnorderedHistoryObservation {
                local_tick,
                before_tick: pair[0].0,
                before_kind,
                after_tick: pair[1].0,
                after_kind,
                compact_history,
            };
            probe.unordered_observations += 1;
            probe.latest_unordered = Some(observation);
        }
    }
}
```

cBournhonesque:
I think I will instead split the PredictionHistory and the ConfirmedHistory to make this easier to reason about; PredictionHistory will be append-only, and ConfirmedHistory can receive updates out of order.
Thanks for raising!

monsterrdev:
> I think I will instead split the PredictionHistory and the ConfirmedHistory to make this easier to reason about; PredictionHistory will be append-only, and ConfirmedHistory can receive updates out of order. Thanks for raising!

That makes sense, I was just in the process of validating an attempt to split it myself. 

If 72c11a702b06a24f6345c2270096015608976401 is close to what you had in mind, I could try to push it over the line. Otherwise we can treat this PR as simply a bug report and close it.

cBournhonesque:
I merged it here! https://github.com/cBournhonesque/lightyear/pull/1507

monsterrdev:
Awesome, thank you

PR #1526, opened 2026-06-23
feat(inputs): opt-in authorize_controlled_targets helper (spoofed-target defense)

mmannerm:
## Summary

Reframes this PR as an **opt-in helper** for the spoofed-target defense, built on the validation seam in #1535 — rather than a compulsory, inline authorization check.

Per the discussion on #1531, `Controlled` / `ControlledBy` are **optional, non-authoritative** helpers (some games legitimately let several clients drive one entity), so lightyear should **not** enforce target authorization by default. Instead it offers a ready-made check a game can switch on:

```rust
app.add_input_validator(authorize_controlled_targets::<MySequence>);
```

## The defense

`receive_input_message` reads `InputTarget::Entity(entity)` off the wire and applies it to that entity's `InputBuffer` with no ownership check — so a modified client can send `InputTarget::Entity(other_player)` to drive an entity it doesn't control.

`authorize_controlled_targets::<S>` is a public `InputSystems::ValidateInputs` system that strips every `InputTarget::Entity` the sender isn't authorized to control (not in its `ControlledByRemote`). The message itself is **kept** even if filtering removes all its targets — an empty `InputMessage` is a legitimate keepalive that still carries `end_tick` and advances confirmation; dropping it would stall the confirmed tick and risk a large rollback. Only the unauthorized targets are removed (before any rebroadcast). Host-client inputs (`is_local`) are exempt; `InputTarget::PreSpawned` (hash-identified) is passed through. It logs (`trace!`) when it strips targets, so a missing `ControlledBy` is diagnosable.

To run your own validation after this helper, order it with `.after(authorize_controlled_targets::<S>)`.

## Tests

- `test_authorize_controlled_targets_helper` — client 0 controls A and forges a target on uncontrolled B; A's authorized input lands (non-overblocking), B's spoofed input is dropped.
- `test_authorize_controlled_targets_keeps_emptied_message` — a fully-unauthorized message is stripped but **kept** (the keepalive isn't dropped); input doesn't land and the pipeline stays healthy.
- `test_authorize_controlled_targets_exempts_host_client` (host-server) — a host-client's inputs are not stripped (`is_local` bypass); verified discriminating.
- `test_authorize_controlled_targets_passes_through_prespawned` — a forged `PreSpawned` target survives the filter.
- `test_custom_validator_ordered_after_authorize` — a custom validator ordered `.after` the helper sees only authorized targets.

## Notes

[...]

PR #1529, opened 2026-06-24
feat(inputs): pluggable validation hook for received input messages

mmannerm:
## Summary

Adds a **pluggable validation hook** for server-side input messages — the
extension point you asked about in #1525 ("a hook that would run on the
`InputMessage`s after deserialization but before they are handled").

Each received `InputMessage<S>` is run through an ordered chain of
user-registered validators **after deserialization and before** it is applied
to any `InputBuffer` (and before rebroadcast). A validator may:

- **mutate** the message — e.g. drop individual `InputTarget`s via
  `inputs.retain(..)`, or
- **reject** it whole — `InputValidation::Reject` skips it entirely.

**No behavior change by default.** lightyear ships no validators here, so the
chain is empty (a no-op) until something is registered. This PR is the
*framework only* — see "Relationship to #1525/#1526" below.

## API

```rust
// implement the trait…
pub trait InputMessageValidator<S: ActionStateSequence>: Send + Sync + 'static {
    fn validate(&self, ctx: &InputValidationContext, msg: &mut InputMessage<S>) -> InputValidation;
    fn name(&self) -> &'static str { core::any::type_name::<Self>() } // for removal
}

// …or just pass a closure (blanket `Fn` impl):
app.add_input_validator::<S>(|ctx, msg| {
    if msg.inputs.len() > 8 { InputValidation::Reject } else { InputValidation::Continue }
});

app.add_input_validators::<S>([Box::new(A), Box::new(B)]); // a group
app.remove_input_validator::<S>(MyValidator::NAME);         // by name
app.clear_input_validators::<S>();                          // all
```

`InputValidationContext` is intentionally small and read-only (validators run
*inside* the receive system, so they can't issue their own queries):
`{ sender, client_id, server_tick }`.

The hook fires once per message in `receive_input_message`, right after the
trace logging and before the interpolation-delay / rebroadcast / apply steps.

## Design options considered

I'd like your steer on the shape before polishing. What this PR implements is
**Option A**.

| | Approach | Pros | Cons |
|---|---|---|---|
| **A (this PR)** | Ordered `Vec<Box<dyn InputMessageValidator<S>>>` resource + `App` ext; closures work via a blanket `Fn` impl | Composable chain; stateful validators; remove/replace by name; closures keep one-off checks terse | Introduces a `Vec<Box<dyn _>>` chain (not used elsewhere in the crate); validators can't run arbitrary ECS queries |
| **B** | A `ValidateInputs` `SystemSet`: stage messages into a resource, let users add **normal systems** that filter/reject before handling | Full `SystemParam`/ECS access (rate-limit on arbitrary components, look at world state); very Bevy-idiomatic | Restructures the drain path; needs a staging resource; ordering/short-circuit semantics are fuzzier across systems |
| **C** | A single `Box<dyn Fn(.., &mut Ctx)>` hook (à la the netcode `on_connect` callback) | Matches an existing lightyear idiom; minimal surface | One hook only — no composition, grouping, or removal |

A and C are "run inside the receive system with a fixed context"; B is "expose
the messages to user systems with full ECS access." They trade conciseness
against power.

## Open question: server-only + thin context, or go generic?

This PR is deliberately **server-only** with a **thin context**. The bigger
design question is whether the framework should be **more generic** — and
whether that would *simplify* the codebase by absorbing checks that are
currently bespoke (see below). Going generic mainly means:

1. giving validators access to the target's `InputBuffer` (state, not just the
   message), and
2. a **client-side** twin for the remote-player input receive path.

I'm inclined toward the more generic form **iff it removes special-cased code**
rather than adding a parallel mechanism. Hence Option B is on the table.

## Refactoring candidates (what could move onto this hook)

Existing, currently-bespoke checks on received input messages that *could*
become validators — with what each would require:

- **`detect_input_history_rewrite` + the late-input-mismatch branch**
  (`crates/inputs/inputs/src/server.rs`). Anomaly detection on inbound inputs;
  today it logs but never rejects. **Requires:** validator access to the
  target's `InputBuffer` (not in the current context).
- **Remote-player input staleness/ordering drop**
  (`crates/inputs/inputs/src/client.rs`, `receive_remote_player_input_messages`
  — drops a message when `last_remote_tick >= end_tick`). Conceptually the same
  "validate an inbound `InputMessage`," but on the **client** receive path.
  **Requires:** a client-side validator chain + `InputBuffer` access.

Explicitly **out of scope** (different domain — should *not* move here):
authority request/response (`replication/.../authority.rs`), `ControlledBy`
visibility (`control.rs`), prespawn hash-collision (`prespawn.rs`),
malformed-packet drops and deserialization errors (message/transport layer).

If we adopt the generic form (buffer access + client side), the two candidates
above could be unified onto one mechanism — that's the "simplifies the
codebase" case for going generic.

## Relationship to #1525 / #1526

This framework is the generalization of the two specific protections; their
merits stay up for review in their own PRs:

- **#1525** (unbounded `end_tick` → memory-exhaustion DoS) becomes an
  `EndTickLookahead` validator. I'll drop the ~10-line validator form in a
  comment below so you can see how it plugs in (it needs only `server_tick`,
  already in the context).
- **#1526** (spoofed `InputTarget::Entity`) becomes a `ControlledByAuthorization`
  validator. It needs the sender's `ControlledByRemote`, which would extend
  `InputValidationContext` with one field.

## Tests

Framework mechanics (no default validators, so input handling is otherwise
unchanged):

- `test_registered_validator_can_reject_input` — a closure that rejects drops the message.
- `test_validator_can_drop_targets_by_mutation` — mutating the message (clearing targets) is honored.
- `test_remove_input_validator_restores_handling` — `remove_input_validator(NAME)` puts handling back.

The full existing input + rollback suites pass unchanged.

### DIFF
--- a/crates/inputs/inputs/src/validation.rs
+++ b/crates/inputs/inputs/src/validation.rs
@@ -1,0 +1,236 @@
+//! Pluggable validation for server-side input messages.
+//!
+//! Input messages arrive untrusted from remote clients. After they are
+//! deserialized and *before* the server applies them to any
+//! [`InputBuffer`](crate::input_buffer::InputBuffer), each message is run
+//! through an ordered chain of [`InputMessageValidator`]s. A validator may
+//!
+//! - **mutate** the message — e.g. drop unwanted [`InputTarget`]s via
+//!   [`InputMessage::inputs`]`.retain(..)`, or
+//! - **reject** the whole message — returning [`InputValidation::Reject`],
+//!   which skips it entirely (no buffer writes, no rebroadcast).
+//!
+//! This module provides *only* the extension point; lightyear ships no
+//! validators by default, so the chain is empty (a no-op) until you register
+//! one. A validator can be any type implementing [`InputMessageValidator`], or
+//! simply a closure (see the blanket impl):
+//!
+//! ```ignore
+//! app.add_input_validator::<MySequence>(|ctx, msg| {
+//!     if msg.inputs.len() > 8 { InputValidation::Reject } else { InputValidation::Continue }
+//! });
+//! ```
+//!
+//! Register, group, replace, or remove validators through
+//! [`InputValidatorAppExt`].
+
+use alloc::boxed::Box;
+use alloc::vec::Vec;
+use bevy_app::App;
+use bevy_ecs::entity::Entity;
+use bevy_ecs::resource::Resource;
+use lightyear_core::id::RemoteId;
+use lightyear_core::tick::Tick;
+
+use crate::input_message::{ActionStateSequence, InputMessage};
+
+/// Read-only context handed to every [`InputMessageValidator`].
+///
+/// Validators run *inside* the input-receive system, so they cannot issue
+/// their own ECS queries; everything they are expected to need is exposed
+/// here. The set of fields is intentionally small and is expected to grow as
+/// new validators need more context (see the module docs and the PR
+/// discussion on scope).
+pub struct InputValidationContext {
+    /// The sender's connection entity on the server (the `ClientOf` / link
+    /// entity the message was received on).
+    pub sender: Entity,
+    /// The sender's network id. `client_id.is_local()` is the host client.
+    pub client_id: RemoteId,
+    /// The server's current tick.
+    pub server_tick: Tick,
+}
+
+/// The outcome of validating a single [`InputMessage`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum InputValidation {
+    /// Keep the (possibly mutated) message and run the next validator.
+    Continue,
+    /// Drop the whole message: stop the chain and skip handling it.
+    Reject,
+}
+
+/// A check applied to each received [`InputMessage`] before it is handled.
+///
+/// Implementations are registered per action-state-sequence type `S` via
+/// [`InputValidatorAppExt`] and stored in an [`InputMessageValidators<S>`]
+/// resource. They run in registration order; the first [`InputValidation::Reject`]
+/// drops the message.
+///
+/// Any matching closure also implements this trait (see the blanket impl
+/// below), so simple one-off checks need no dedicated type.
+pub trait InputMessageValidator<S: ActionStateSequence>: Send + Sync + 'static {
+    /// Inspect (and optionally mutate) `message`. Return
+    /// [`InputValidation::Reject`] to drop it.
+    fn validate(
+        &self,
+        ctx: &InputValidationContext,
+        message: &mut InputMessage<S>,
+    ) -> InputValidation;
+
+    /// A stable identifier used to remove or replace this validator (see
+    /// [`InputValidatorAppExt::remove_input_validator`]). Defaults to the
+    /// concrete type name.
+    fn name(&self) -> &'static str {
+        core::any::type_name::<Self>()
+    }
+}
+
+/// Any `Fn(&InputValidationContext, &mut InputMessage<S>) -> InputValidation`
+/// is an [`InputMessageValidator`], so one-off checks can be registered as a
+/// closure without defining a type.
+///
+/// Closures take the default [`name`](InputMessageValidator::name) (their
+/// opaque type name), so they are not practical to remove by name — define a
+/// named type if you need [`InputValidatorAppExt::remove_input_validator`].
+impl<S, F> InputMessageValidator<S> for F
+where
+    S: ActionStateSequence,
+    F: Fn(&InputValidationContext, &mut InputMessage<S>) -> InputValidation
+        + Send
+        + Sync
+        + 'static,
+{
+    fn validate(
+        &self,
+        ctx: &InputValidationContext,
+        message: &mut InputMessage<S>,
+    ) -> InputValidation {
+        self(ctx, message)
+    }
+}
+
+/// Resource holding the ordered validator chain for [`InputMessage<S>`].
+///
+/// Inserted (empty) by `ServerInputPlugin`; mutate it through
+/// [`InputValidatorAppExt`].
+#[derive(Resource)]
+pub struct InputMessageValidators<S: ActionStateSequence> {
+    validators: Vec<Box<dyn InputMessageValidator<S>>>,
+}
+
+impl<S: ActionStateSequence> Default for InputMessageValidators<S> {
+    fn default() -> Self {
+        Self {
+            validators: Vec::new(),
+        }
+    }
+}
+
+impl<S: ActionStateSequence> InputMessageValidators<S> {
+    /// Append a single validator (a type or a closure).
+    pub fn push(&mut self, validator: impl InputMessageValidator<S>) {
+        self.validators.push(Box::new(validator));
+    }
+
+    /// Append several already-boxed validators (a group).
+    pub fn extend(
+        &mut self,
+        validators: impl IntoIterator<Item = Box<dyn InputMessageValidator<S>>>,
+    ) {
+        self.validators.extend(validators);
+    }
+
+    /// Remove every validator.
+    pub fn clear(&mut self) {
+        self.validators.clear();
+    }
+
+    /// Remove every validator whose [`InputMessageValidator::name`] matches.
+    pub fn remove(&mut self, name: &str) {
+        self.validators.retain(|v| v.name() != name);
+    }
+
+    /// Number of registered validators.
+    pub fn len(&self) -> usize {
+        self.validators.len()
+    }
+
+    /// Whether the chain is empty.
+    pub fn is_empty(&self) -> bool {
+        self.validators.is_empty()
+    }
+
+    /// Run the chain over `message`. Returns `false` if any validator rejected
+    /// it (the caller should then skip the message).
+    pub fn validate(
+        &self,
+        ctx: &InputValidationContext,
+        message: &mut InputMessage<S>,
+    ) -> bool {
+        for validator in &self.validators {
+            if validator.validate(ctx, message) == InputValidation::Reject {
+                return false;
+            }
+        }
+        true
+    }
+}
+
+/// App-builder extension for registering, grouping, and removing
+/// [`InputMessageValidator`]s.
+pub trait InputValidatorAppExt {
+    /// Append one validator (type or closure) to the chain for `S`.
+    fn add_input_validator<S: ActionStateSequence>(
+        &mut self,
+        validator: impl InputMessageValidator<S>,
+    ) -> &mut Self;
+
+    /// Append a group of already-boxed validators to the chain for `S`.
+    fn add_input_validators<S: ActionStateSequence>(
+        &mut self,
+        validators: impl IntoIterator<Item = Box<dyn InputMessageValidator<S>>>,
+    ) -> &mut Self;
+
+    /// Remove all validators for `S`.
+    fn clear_input_validators<S: ActionStateSequence>(&mut self) -> &mut Self;
+
+    /// Remove the validator(s) named `name` for `S`.
+    fn remove_input_validator<S: ActionStateSequence>(&mut self, name: &str) -> &mut Self;
+}
+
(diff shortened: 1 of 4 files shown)


mmannerm:
For concreteness, here's how the #1525 DoS protection becomes a built-in validator on this hook — it needs only `server_tick`, which is already in `InputValidationContext`:

```rust
/// Rejects messages whose `end_tick` is implausibly far from the server tick,
/// so a forged far-future `end_tick` can't drive `InputBuffer::set_raw` to
/// allocate one entry per intermediate tick (memory-exhaustion DoS).
pub struct EndTickLookahead {
    pub max_ahead: i32,  // default 64  (~1s @ 64Hz; legit input delay is 0–3 ticks)
    pub max_behind: i32, // default 256 (also rejects the i32::MIN wrap-around at server+2^31)
}

impl EndTickLookahead {
    pub const NAME: &'static str = "lightyear_inputs::EndTickLookahead";
}

impl<S: ActionStateSequence> InputMessageValidator<S> for EndTickLookahead {
    fn validate(&self, ctx: &InputValidationContext, msg: &mut InputMessage<S>) -> InputValidation {
        let delta = msg.end_tick - ctx.server_tick; // Tick - Tick -> i32
        if (-self.max_behind..=self.max_ahead).contains(&delta) {
            InputValidation::Continue
        } else {
            InputValidation::Reject
        }
    }
    fn name(&self) -> &'static str { Self::NAME }
}
```

Registration (the maintainer decides whether it's default-on or opt-in):

```rust
app.add_input_validator::<S>(EndTickLookahead::default());
// downstream can opt out:
app.remove_input_validator::<S>(EndTickLookahead::NAME);
```

The spoofed-target defense (#1526) is the same idea but mutating: it `retain`s
only `InputTarget::Entity` values the sender controls. That one needs the
sender's `ControlledByRemote`, i.e. it would add one field to
`InputValidationContext` — which is the smallest concrete example of the
"thin vs richer context" question in the description.

So the open call is really: keep this server-only with a minimal context and
land the two protections as built-in validators on top, or go for the more
generic form (buffer access + a client-side chain) if it lets us also retire
the bespoke `detect_input_history_rewrite` / remote-input-staleness checks.
Happy to push whichever direction you prefer.

mmannerm:
Pushed a second commit (`buffer-aware validation context + ergonomics`) addressing the main gap in the first cut, and I want your steer on two design questions before going further.

**What changed**
- `InputValidationContext` is now a borrowing, `#[non_exhaustive]` type that exposes a **read-only per-target `InputBuffer` accessor** (`InputBufferProvider`), built from the receive system's existing buffer query — so buffer-aware checks are expressible **without** giving validators arbitrary ECS access or introducing a new system. (Resolves `InputTarget::Entity`; `PreSpawned` returns `None` for now.)
- `named(name, validator)` so closures can be removed via `remove_input_validator` (bare closures share an opaque `name()`).
- The chain reports which validator rejected; the receive path now emits a structured `server_input_message_rejected` event instead of a bare `trace!`.
- Documented that host-client / rebroadcast-disabled inputs return before the chain and so aren't seen by validators.

**Q1 — new system, or in-system? (I kept it in-system.)** Validation has to run at one fixed point — after deserialize, before rebroadcast and before the buffer write — and abort the message with an early return. A separate `ValidateInputs` SystemSet + staging resource can only *approximate* that position, would retain attacker-controlled messages before validation runs, and (the real concern) would expose the rollback-feeding `earliest_mismatch_input` aggregation — ordered `.before(RollbackSystems::Check)` — to downstream `configure_sets` reordering. An inline chain has a guaranteed, non-reorderable position. If a validator ever genuinely needs full ECS access, the existing `InputSystems::ReceiveInputs` set is already there to order a normal system around, additively. Does that match your intent, or do you want the SystemSet shape?

**Q2 — how generic? (server-only today; proposing a measured expansion.)** With rebroadcast on, the server relays the (possibly validator-mutated) message to peers while the client remote-player path has no chain, so the peer-facing surface is currently unvalidated. I'd like to add a **separate client-side chain** (its own resource + app-ext) wired into `receive_remote_player_input_messages`, and move the client staleness drop (`last_remote_tick >= end_tick`) onto it as a default validator — that's a real accept/reject gate and a clean fit. I'd **leave `detect_input_history_rewrite` inline** (it's observability — it logs and always continues, not an accept/reject decision), and **leave the late-input-mismatch check inline** (it's fused into the `&mut update_buffer` write, so a pre-write read-only validator can't reproduce it without `&mut` buffer access). Happy to go more generic if you'd prefer, but that's the line I'd draw to keep the hook honest.

If Q1/Q2 look right I'll add the client-side twin next. The two specific protections from #1525/#1526 layer on top of this as built-in validators (DoS needs only `server_tick`; spoof adds the sender's `ControlledByRemote` to the context).

🤖 Co-authored with [Claude Code](https://claude.com/claude-code) (Claude Opus 4.8)

crates/inputs/inputs/src/validation.rs
```diff
@@ -0,0 +1,320 @@
...
+#[non_exhaustive]
+pub struct InputValidationContext<'a, S: ActionStateSequence> {
+    /// The sender's connection entity on the server (the `ClientOf` / link
+    /// entity the message was received on).
+    pub sender: Entity,
+    /// The sender's network id. `client_id.is_local()` is the host client.
+    pub client_id: RemoteId,
+    /// The server's current tick.
+    pub server_tick: Tick,
+    buffers: &'a dyn InputBufferProvider<S>,
```
cBournhonesque, line 73: Do we need the trait object here? We could just make the InputValidationContext generic on the InputBufferProvider, no?

cBournhonesque:
Some thoughts:
- I don't think it is that easy to craft fake input messages because of the protections given by netcode (https://github.com/mas-bandwidth/netcode/blob/main/STANDARD.md) but this seems good to have

- in general we WILL need some sort of ECS access in the input validation; simply because a lot of anti-cheats do some coarse 'is this plausible' check. For example I believe Battlefield does client-side hit registration (to avoid dealing with lag compensation) + some validation logic that checks if the 'hit detected' is plausible.
As you said, the user is already free to add a system in-between `MessageSystems::Receive` and `InputSystems::Buffer` to do the validation themselves; I guess we just have to provide an example?

- The separate system might be expensive and not needed in a lot of cases, so it's probably good to also have the in-system validation (this PR) for simpler checks that don't require ECS access. I think adding a client-side validator is really not a priority at this point, so let's just focus on the server one. The issue is that without ECS access, there's only a limited amount of validation that you can do (basically only https://github.com/cBournhonesque/lightyear/pull/1525) so maybe it's worth just merging #1525 instead of implementing a full in-system validator system? What do you think?

mmannerm:
Really useful framing — I think the right way to resolve this is to nail down **what lightyear owns vs. what the game owns** in the input pipeline, because that line decides what should merge and what shouldn't exist.

## Threat model: netcode is necessary but not sufficient

netcode secures the **connection**: encryption, connect-token auth, replay protection, packet-level DoS. It guarantees a packet came from *an authenticated client* and wasn't forged or replayed *on the wire*.

It says nothing about the **application-level content** that an authenticated client sends. The standard cheat model is a player who controls their own machine and runs a modified client — still fully authenticated. Such a client can put anything in an `InputMessage`:

- `end_tick = server_tick + 30_000` → unbounded `InputBuffer` allocation (#1525). Authentic connection, malicious payload.
- `InputTarget::Entity(other_player)` → inputs routed to a victim's buffer (#1526). Authentic connection, malicious *target*.

So "netcode makes fake messages hard" conflates *wire forgery* (netcode stops it) with *malicious-but-authentic content* (it can't). Both PRs defend against the authenticated-modified-client case, which is exactly the case netcode's STANDARD leaves to the application.

## The ownership boundary

The clean split, I think, is **provenance & transport-integrity (lightyear)** vs **semantic plausibility (game)**:

- **lightyear owns provenance + the integrity of its own input plumbing.** Only lightyear knows the connection↔entity ownership relationship (`ControlledBy`/`ControlledByRemote`) and the sender identity, and only lightyear demuxes an `InputMessage` onto per-entity `InputBuffer`s. So "is this sender allowed to drive this target?" (#1526) and "is this message bounded enough not to blow up my buffer?" (#1525) can *only* be enforced here. They're game-agnostic — true for every lightyear game.
- **the game owns semantic plausibility.** Given an input that provably came from the right client and is well-formed, *is the action legal under game rules?* (speed/cooldown/hit-plausibility…). That needs full game state and game-specific policy; lightyear can't and shouldn't own it.

Concrete illustration from our project (rauta): we already sanitize client inputs server-side — a system that clamps `Move`/`Look` to the unit-vector contract and rejects non-finite values (anti-speedhack / NaN-poisoning), running before any authoritative consumer. But that system reads `ActionState` **after** lightyear has already routed the input to the entity — it has no way to ask "did the authorized client send this?". The value-sanitization is squarely ours; the provenance check is structurally lightyear's. #1525/#1526 are the part we can't do from the game side.

## Proposal

1. **Merge #1525 and #1526.** Both sit on lightyear's side of the line: a transport-integrity bound and a provenance check. Happy to defend #1526 specifically — it's an *authorization* gap (the receive path writes to whatever entity the message names, with no owner check), not a forgery problem netcode covers.
2. **Close #1529.** I'm convinced the in-system *validator framework* is the wrong middle abstraction here: the two checks above are ~15-line functions, not a registry; the codebase has no other `Box<dyn>` chains; and the genuinely valuable validation (plausibility/anti-cheat) needs full ECS access, which is a different mechanism entirely (below). I'd rather not maintain a framework whose only realistic occupants are the two security fixes.

## How games should validate their own inputs (and where a "generic" API sits)

For the semantic tier, the idiomatic answer is a **normal Bevy system with full ECS access at a known pipeline slot** — and I think that's where a *general* validation API actually belongs (distinct from the two security fixes above). A concrete shape we like for it:

```
MessageSystems::Receive → StageReceivedInputs → ValidateReceivedInputs → BufferInputs
```

lightyear stages the decoded inputs, exposes **`ValidateReceivedInputs`** as a public `SystemSet` where game systems (full `SystemParam` access) inspect / clamp / reject them, then **`BufferInputs`** applies what survives. The key ordering point that resolves my earlier worry about staging: the **transport-integrity bound (#1525) runs before staging** (you never stage an unbounded message) and **provenance (#1526) before validation**, so by the time game validation runs it operates on inputs that are already bounded *and* already authorized. That keeps the security guarantees inline and non-reorderable, while giving plausibility checks a clean, ECS-capable seam. (One detail to design carefully: the rollback-feeding mismatch aggregation that's ordered `.before(RollbackSystems::Check)` — the staged pipeline needs to preserve that.)

On "can it be generic?":

- *Transport-level* validation generalizes — that's #1525/#1526, game-agnostic, lightyear's.
- *Plausibility/anti-cheat* doesn't fully generalize — it's game policy (rules, physics, abilities). What's reusable is the **pattern and the slot**, not the logic: (a) prefer server-authoritative simulation so most "cheats" are simply clamped by the sim; (b) for the rest, a system in `ValidateReceivedInputs` with game state. A generic anti-cheat layer would be a game-side crate on top of lightyear's provenance guarantees + this slot.

I'll **open an RFC issue** to design that staged pipeline properly (the `Stage`/`Validate`/`Buffer` sets, the rollback-ordering constraint, and a worked example based on the rauta clamp/finite-check). That's the right venue for the ECS-access validation you're describing — separate from landing the two security fixes here.

🤖 Co-authored with [Claude Code](https://claude.com/claude-code) (Claude Opus 4.8)

mmannerm:
Closing in favor of landing the two transport-layer fixes directly (#1525, #1526) and designing the game-side / ECS-access validation as a staged `Receive → Stage → Validate → Buffer` pipeline in a forthcoming RFC, per the discussion above. The framework here was the wrong middle abstraction. Thanks for the steer!

🤖 Co-authored with [Claude Code](https://claude.com/claude-code) (Claude Opus 4.8)

mmannerm:
Re-opened so that you can close it after reading it

PR #1547, opened 2026-07-02
Support Gilbert-Elliott packet-loss model

### LINKED ISSUE #1546: Support Gilbert-Elliot model for packet loss simulation
Lightyear should support simulating packet drops using the [Gilbert–Elliott model](https://en.wikipedia.org/wiki/Burst_error#Gilbert%E2%80%93Elliott_model) as it simulates real packet drop behavior more accurately.

Lightyear's current method to simulate packet drops is to follow a [Bernoulli process](https://en.wikipedia.org/wiki/Bernoulli_process):

> The chance of a packet drop is based on a fixed probability that is independent of any previous packet drops.

Real network packet drops do not follow this model. Instead, real network packet drops are correlated, so they tend to drop in bursts. The Gilbert–Elliott model simulates this bursty behavior. The model holds onto a "good" and "bad" state.

> If we are in a bad state, packet loss probability is high.
>
> If we are in a good state, packet loss probability is low.
> 
> There is a separate probability of switching from the good to bad state.
> 
> There is a separate probability of switching from the bad to good state.

I think this new method should completely replace the current method. We don't lose functionality because the new method can be configured to simulate the current method (the Gilbert–Elliott model is a generalization of the Bernoulli process).

nick-e:
Replace `LinkConditioner`'s per-packet drop rate with a two-state Gilbert-Elliott Markov model so simulated packet loss occurs in correlated bursts, matching real packet-loss behavior. The conditioner now carries persistent good/bad loss state across packets.

Create a `PacketLossConfig` struct to configure the parameters of the Gilbert-Elliott model. `PacketLossConfig::fixed_loss_probability()` allows users to use the original per-packet loss probability method. `PacketLossConfig::bursty()` builds a burst model from a target loss rate and burst length.

Closes #1546.

### DIFF
--- a/crates/io/link/src/conditioner.rs
+++ b/crates/io/link/src/conditioner.rs
@@ -6,6 +6,221 @@ use lightyear_core::time::Instant;
 use lightyear_utils::ready_buffer::ReadyBuffer;
 use rand::RngExt;
 
+/// Configuration for the probability of network packet loss.
+///
+/// We use the [Gilbert–Elliott model] in order to simulate realistic network
+/// behavior which is bursty in nature. The model keeps track of a hidden "good"
+/// and "bad" state:
+///
+/// * If we are in a bad state, packet loss probability is high
+/// * If we are in a good state, packet loss probability is low
+/// * We switch between good and bad states based on probabilities unrelated to
+///   whether or not a packet was lost.
+///
+/// # Choosing a constructor
+///
+/// Each constructor frees more of the four probabilities and can reproduce
+/// one more kind of real-world loss pattern.
+///
+/// * [`fixed_loss_probability`](Self::fixed_loss_probability): The probability
+///   of loss for every packet is the same. Use when only the overall rate
+///   matters. At low rates it essentially never drops consecutive packets, so
+///   it under-stresses anything tuned for burst tolerance (input redundancy,
+///   interpolation buffers).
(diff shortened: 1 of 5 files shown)


crates/io/link/src/conditioner.rs
```diff
@@ -120,7 +272,7 @@ impl LinkConditionerConfig {
         LinkConditionerConfig {
             incoming_latency: Duration::from_millis(40),
             incoming_jitter: Duration::from_millis(6),
-            incoming_loss: 0.002,
+            incoming_loss: PacketLossConfig::fixed_loss_probability(0.002),
```
cBournhonesque, line 368: I would have expected these to be replaced with the `bursty()` model

nick-e: `bursty()` is now renamed to `simple_gilbert()`.

I didn't change the values because I don't know what settings to pick for `simple_gilbert()` yet. I'm working on capturing real network traffic of connections within a city and across USA's coasts in order to create realistic configs.

nick-e: As a side note, it would be cool to have a `LinkConditionerConfig::fit()` function that takes in a sample of real network data and calculates a `LinkConditionerConfig` based on it. It's something I'm looking in to.

nick-e:
Closing in favor of #1550 because this PR accidentally follows the wrong branch of the git fork and you can't change it apparently.

PR #1550, opened 2026-07-02
Support Gilbert-Elliott packet-loss model

### LINKED ISSUE #1546: Support Gilbert-Elliot model for packet loss simulation

nick-e:
Closes https://github.com/cBournhonesque/lightyear/issues/1546.

Replace LinkConditioner's per-packet drop rate with a two-state Gilbert-Elliott Markov model so simulated packet loss occurs in correlated bursts, matching real packet-loss behavior.

Give `LinkConditioner` a good/bad state. A good state denotes the link is healthy and not dropping packets. A bad state denotes the link is dropping a lot of packets. It also opens up the possibility of adding an extra latency to packets during a bad state.

Add builder methods to `LinkConditionerConfig` to configure the packet loss model.

### DIFF
--- a/crates/io/link/Cargo.toml
+++ b/crates/io/link/Cargo.toml
@@ -28,6 +28,9 @@ bevy_ecs.workspace = true
 bevy_reflect.workspace = true
 bevy_utils.workspace = true
 
+[dev-dependencies]
+approx.workspace = true
+
 [lints]
 workspace = true
 
--- a/crates/io/link/src/conditioner.rs
+++ b/crates/io/link/src/conditioner.rs
@@ -6,17 +6,82 @@ use lightyear_core::time::Instant;
 use lightyear_utils::ready_buffer::ReadyBuffer;
 use rand::RngExt;
 
+/// The simulated link's current condition. Used as the good/bad state in the
+/// Gilbert–Elliott model for determining probability of packet loss (See
+/// [`LinkConditionerConfig`]).
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum LinkConditionerState {
+    /// The link is healthy. Usually means packets are not be getting lost or
+    /// delayed.
+    #[default]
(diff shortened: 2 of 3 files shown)


nick-e:
To address the convo in the previous PR:

I didn't change the values for `good_condition()`, `average_condition()`, and `poor_condition()` because I don't know what settings to pick yet. I'm working on capturing real network traffic of connections within a city and across USA's coasts in order to create realistic configs. I need #1549 for that.

nick-e:
> To address the convo in the previous PR:
> 
> I didn't change the values for `good_condition()`, `average_condition()`, and `poor_condition()` because I don't know what settings to pick yet. I'm working on capturing real network traffic of connections within a city and across USA's coasts in order to create realistic configs. I need #1549 for that.

I've added some `LinkConditionerConfig` presets in #1562 (that PR contains the `fit()` function used to create the presets).

cBournhonesque:
Thanks, looks great

PR #1562, opened 2026-07-05
Support fitting `LinkConditionerConfig` over real network data

### LINKED ISSUE #1577: Support fitting `LinkConditionerConfig` to real network data
Add the ability to configure a `LinkConditionerConfig` to simulate the network behavior/patterns of a given capture of real lightyear network traffic. I believe this would look something along the lines of:

```rust
struct Packet {
  id: usize,

  /// `None` if lost.
  rtt: Option<Duration>
}

impl LinkConditionerConfig {
  /// Returns a config that simulates the network behavior derived from `packets`.
  fn fit(packets: &[Packet]) -> Self {
    ...
  }
}
```

If lightyear decides to implement the Gilbert-Elliott model for packet loss (#1546) then this `fit()` function will make it easier to construct useful conditioner configs as the Gilbert-Elliott model can be confusing to setup manually. Users would instead record network traffic during a real lightyear connection and then feed it into `fit()`.

I already have a working implementation to fit a config that uses the Gilbert-Elliott model.

nick-e:
> [!NOTE]
> This PR is stacked on top of #1550 because it is a dependency. The first commit of this PR is actually #1550. Only the second commit is new here.

Closes https://github.com/cBournhonesque/lightyear/issues/1577.

Add `LinkConditionerConfig::fit()`. It constructs a conditioner config using a provided capture of packets sent by a link. This capture includes packet RTT, ID, and if the packet got lost. The conditioner will try to replicate packet-loss probabilities and latency seen in the capture.

Create the `RttOverheadEstimatorPlugin` plugin. It can provide an estimate of the RTT overhead introduced by lightyear. This estimate is needed for `fit()`.

Create `LinkConditionerConfig` presets that simulate network conditions between some US cities.

### DIFF
--- a/crates/transport/transport/src/rtt_overhead.rs
+++ b/crates/transport/transport/src/rtt_overhead.rs
@@ -1,0 +1,341 @@
+//! Real time estimation of the packet RTT (round-trip time) overhead introduced
+//! by lightyear. Include [`RttOverheadEstimatorPlugin`] in your app and trigger
+//! [`EstimateRttOverhead`] to calculate an estimation. The estimation is
+//! published once as an [`RttOverheadEstimated`] event. You can use this as
+//! the RTT estimate required by
+//! [`LinkConditionerConfig::fit`](lightyear_link::prelude::LinkConditionerConfig::fit).
+//!
+//! RTT overhead is comprised of:
+//!
+//! * The delay between lightyear receiving a packet and lightyear sending the
+//!   ack for it (the ack is not sent until the next scheduled outgoing packet).
+//! * The delay between lightyear receiving an ack and lightyear actually
+//!   processing it.
+//!
+//! To estimate the RTT overhead, the estimator spawns two endpoints (marked
+//! with [`EstimatorEndpoint`]), connects them over an in-process
+//! [`CrossbeamIo`] pair, and sends packets between them through lightyear until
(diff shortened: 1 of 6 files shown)


crates/transport/transport/src/rtt_overhead.rs
```diff
@@ -0,0 +1,341 @@
...
+//! [`LinkConditionerConfig::fit`](lightyear_link::prelude::LinkConditionerConfig::fit).
+//!
+//! RTT overhead is comprised of:
+//!
+//! * The delay between lightyear receiving a packet and lightyear sending the
+//!   ack for it (the ack is not sent until the next scheduled outgoing packet).
+//! * The delay between lightyear receiving an ack and lightyear actually
+//!   processing it.
+//!
+//! To estimate the RTT overhead, the estimator spawns two endpoints (marked
```
cBournhonesque, line 15: The RTT overhead would be different on client vs server since they have different frame-rate.
Should the user run this plugin on both and then add up both overheads?

Also to be clear we don't consider delays based on replication rate since the RTT is estimated using PingChannel and both peers answer pings in the same frame they have received it, so we only need to measure delays tied to frames.

I guess it also doesn't measure any delay that could be introduced by the transport itself (WebTransport congestion control, etc.)

PR #1576, opened 2026-07-07
fix: bei host client input handling

Ploruto:
This simply removes a block in the `server.rs` in the bei example:
```rs
if is_host_server {
        if let Ok(controlled_by) = controlled_by.get(trigger.context) {
            if host_clients.get(controlled_by.owner).is_ok() {
                return;
            }
        }
    }
```
This would lead to the host client never firing the shared movement logic on it's own local input.

It works as I expected after applying those changes.

### DIFF
--- a/examples/bevy_enhanced_inputs/src/server.rs
+++ b/examples/bevy_enhanced_inputs/src/server.rs
@@ -121,13 +121,7 @@ fn movement(
     if is_host_server && !server_actions.contains(trigger.action) {
         return;
     }
-    if is_host_server {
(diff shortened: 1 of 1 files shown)


cBournhonesque:
Thanks. Your PR works because we need the inputs to trigger either on the client or the server, for host-server mode, not both.
The inputs are already excluded from the client here: https://github.com/cBournhonesque/lightyear/blob/main/examples/bevy_enhanced_inputs/src/client.rs#L54

PR #1579, opened 2026-07-08
Emit `PacketAckedLate` when an ack arrives for a lost packet

### LINKED ISSUE #1578: Support packet-acked-late event
Lightyear should fire a `PacketAckedLate` event when it receives an ack for a packet it already marked as lost in order to create more accurate `LinkConditionerConfig`s.

Before I go further, here are some definitions since the terms are otherwise ambiguous:

* **Packet Lost** - lightyear did not receive an ack within the nack timeout (`rtt_estimate * 1.5`) after sending the packet. This is what `PacketLost` reports.
* **Packet Dropped** - the network actually discarded the packet (or its ack) so it never reached its destination.

"Dropped" isn't official terminology, it's just a way to differentiate from "lost". I realize that `LinkConditionerConfig`'s fields and comments say "loss" but they actually describe what I'm calling "dropped" here.

A **dropped** packet always becomes a **lost** packet, but a **lost** packet was not necessarily **dropped**. The packet's ack may just have arrived after the nack timeout.

This is an issue when trying to analyze a capture of real lightyear packet traffic data in order to create a `LinkConditionerConfig`. We currently only have access to `PacketLost` and `PacketAcked` and so we have to consider all **lost** packets as **dropped** packets. This results in:

* Overestimating how many packets actually get **dropped** (packet-drop probability becomes too high)
* Underestimating packet latency because we don't get the RTTs of the packets with late acks (packet latency becomes too low)

A `PacketAckedLate` event would tell us that a **lost** packet didn't actually get **dropped** thus we can create a more accurate conditioner config.

nick-e:
Closes #1578.

Fire the `PacketAckedLate` event when a packet that lightyear had already declared lost receives an acknowledgement. The packet is still considered lost. The event is for more accurate monitoring of network conditions.

Don't retain lost packet meta info forever. There is still a max wait time for lost packets to be late-acked before lightyear stops tracking them all together.

Add a diagnostic for late-acked packets.

### DIFF
--- a/crates/transport/transport/src/packet/header.rs
+++ b/crates/transport/transport/src/packet/header.rs
@@ -120,6 +120,14 @@ const MIN_NACK_MILLIS: u64 = 10;
 /// maximum number of seconds after which we consider a packet lost
 const MAX_NACK_SECONDS: u64 = 3;
 
+/// How long a lost packet id (with its send time) is retained so that a late
+/// acknowledgment can still be matched to it and reported as late-acked.
+///
+/// Measured from the packet's send time. The nack timeout is capped at
+/// [`MAX_NACK_SECONDS`], so this guarantees at least `MAX_NACK_SECONDS` of
+/// retention after the packet was reported lost.
+const LATE_ACK_RETENTION: Duration = Duration::from_secs(2 * MAX_NACK_SECONDS);
+
 /// Keeps track of sent and received packets to be able to write the packet headers correctly
 /// For more information: [GafferOnGames](https://gafferongames.com/post/reliability_ordering_and_congestion_avoidance_over_udp/)
 #[derive(Debug)]
@@ -131,8 +139,21 @@ pub struct PacketHeaderManager {
     // so we can resend them when dropped
     // sent_packets_not_acked: HashSet<PacketId>,
     sent_packets_not_acked: IndexMap<PacketId, Duration, FixedHasher>,
+
+    /// Send times of the packets that were reported lost (nacked) in case a
+    /// late ack does arrive. Entries are held on to until
+    /// [`LATE_ACK_RETENTION`] has passed since their send time.
+    ///
+    /// Packets will still be considered lost even if a late ack does arrive.
+    recently_lost: IndexMap<PacketId, Duration, FixedHasher>,
+
     stats_manager: PacketStatsManager,
     pub(crate) lost_packets: Vec<PacketId>,
+
+    /// RTTs of the packets that were reported lost but whose ack arrived
+    /// afterwards.
+    pub(crate) late_acked_packets: Vec<(PacketId, Duration)>,
+
     pub(crate) newly_acked_packets: IndexMap<PacketId, Duration, FixedHasher>,
 
     // keep track of the packets that were received (last packet received and the
@@ -157,7 +178,9 @@ impl PacketHeaderManager {
             next_packet_id: PacketId(0),
             stats_manager: PacketStatsManager::default(),
             sent_packets_not_acked: IndexMap::default(),
+            recently_lost: IndexMap::default(),
             lost_packets: vec![],
+            late_acked_packets: vec![],
             newly_acked_packets: IndexMap::default(),
             recv_buffer: ReceiveBuffer::new(),
             nack_rtt_multiple,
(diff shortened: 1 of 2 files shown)


crates/transport/transport/src/packet/header.rs
```diff
@@ -120,6 +120,14 @@ const MIN_NACK_MILLIS: u64 = 10;
 /// maximum number of seconds after which we consider a packet lost
 const MAX_NACK_SECONDS: u64 = 3;
 
+/// How long a lost packet id (with its send time) is retained so that a late
+/// acknowledgment can still be matched to it and reported as late-acked.
+///
+/// Measured from the packet's send time. The nack timeout is capped at
+/// [`MAX_NACK_SECONDS`], so this guarantees at least `MAX_NACK_SECONDS` of
+/// retention after the packet was reported lost.
+const LATE_ACK_RETENTION: Duration = Duration::from_secs(2 * MAX_NACK_SECONDS);
```
cBournhonesque, line 129: I don't really see the point of this; we could just double the MAX_NACK_SECONDS then. But you are worried that this could have some adverse effects since we might depend on nacks for re-sending replication data, etc.?

3s seems pretty long, how many cases do you think we would have where we still receive the packet between [3, 6]s?

nick-e: Are you asking if late-acks are worth tracking or why `LATE_ACK_RETENTION` is double `MAX_NACK_SECONDS`?

If it's the first question, `MAX_NACK_SECONDS` is the max that a nack timeout can be. It is not how long every nack timeout waits before considering a packet lost. Nack timeouts are usually set much lower to the point that an ack could reasonably miss the timeout. In other words, a late-ack could occur well within the `MAX_NACK_SECONDS` window.

If it's the second question, the nack timeout is capped at `MAX_NACK_SECONDS`, and retention is measured from send time, so if `LATE_ACK_RETENTION` were set to `MAX_NACK_SECONDS`, a packet nacked at the cap would be evicted the moment it was declared lost. `LATE_ACK_RETENTION` needs extra time to cover how much later the ack can still plausibly arrive. So at minimum `LATE_ACK_RETENTION > MAX_NACK_SECONDS`.

Now the next question is "how much extra time should we wait?" That is a bit subjective. Lightyear estimates how long an ack can plausibly take with `1.5 * the link's RTT`, capped at `MAX_NACK_SECONDS`. The cap is at RTT = 2s (1.5 * 2s = 3s), which is the slowest link lightyear operates on with its intended margin. On that link, `LATE_ACK_RETENTION = 2 * MAX_NACK_SECONDS` accepts a total ack delay of 6s which is three times the RTT. Any ack slower than that is so statistically unlikely that it does not improve our statistical analysis of the network conditions.

crates/transport/transport/src/packet/header.rs
```diff
@@ -131,8 +139,21 @@ pub struct PacketHeaderManager {
     // so we can resend them when dropped
     // sent_packets_not_acked: HashSet<PacketId>,
     sent_packets_not_acked: IndexMap<PacketId, Duration, FixedHasher>,
+
+    /// Send times of the packets that were reported lost (nacked) in case a
+    /// late ack does arrive. Entries are held on to until
+    /// [`LATE_ACK_RETENTION`] has passed since their send time.
+    ///
+    /// Packets will still be considered lost even if a late ack does arrive.
+    recently_lost: IndexMap<PacketId, Duration, FixedHasher>,
```
cBournhonesque, line 148: I'm a bit reluctant to introduce these new data-structures + tracking for a feature that is fairly niche. Would it be possible to make all this optional?

crates/transport/transport/src/packet/header.rs
cBournhonesque: I guess my MAX_NACK_SECONDS was already extremely generous; anything above 500ms RTT probably would not happen. I was jus thinking that 6s to wait for an ack seemed a bit excessive. Maybe we should reduce the values to 1.5s and 3s?

crates/transport/transport/src/plugin.rs
```diff
@@ -63,12 +63,35 @@ pub struct PacketAcked {
...
+/// receiver. A [`PacketAckedLate`] event could fire afterwards if the
+/// acknowledgement for the packet does finally arrive.
 #[derive(EntityEvent)]
 pub struct PacketLost {
     pub entity: Entity,
     pub packet_id: u32,
 }
 
+/// Event triggered on a [`Transport`] entity when an acknowledgement arrives
+/// for a packet that was already reported as [`PacketLost`].
```
cBournhonesque, line 77: I guess it would be interesting to see how often this happens.
The current strategy of declaring a packet lost after 1.5 RTT is slightly risky since there is no guarantee that the remote would send packets frequently enough for us to receive an ACK in time. The server mostly sends packets only for pings and for replication, and replication could happen at a fairly low frequency, which means the client might take a while before receiving an ACK.

PR #1584, opened 2026-07-09
Send UDP packets in post-update

nick-e:
Make the UDP transport send its packet in post-update instead of pre-update. Packets will no longer have to wait for the next frame before being sent out. This reduces average and variance of packet RTTs resulting in less latency and less rollbacks. Server-UDP, crossbeam, and aeronet transports already do this.


cBournhonesque:
Nice catch, thanks!

PR #1594, opened 2026-07-14
add late join single replication target test

hlbarber:
Asserts that new clients DONT get components replicated to them if the target is `Single` and not pointing at them.


crates/tests/src/client_server/replication.rs
hlbarber, line 144: This is the prior art.

Issue #1511, opened 2026-06-16
Predicted component wrongly removed on state rollback when no history sample exists at-or-before the rollback tick (component sibling of #1467)


stevesloan:
## Summary

A **predicted component that is continuously present** in the world is **removed** from the predicted entity during a state rollback whenever neither its `PredictionHistory<C>` nor its `ConfirmedHistory<C>` has a sample at-or-before the rollback tick.

This is the same misinterpretation that #1467 ("Rollback resources incorrectly removed after sync+rollback") fixed for **resources** — but the fix was never applied to the **component** path. It bites components that change rarely (or once, then never): a near-constant component's only sample gets shifted past the rollback target by a sync `TickEvent`, or pruned out of the retained window, after which `prepare_rollback` reads `None` and treats it as a removal.

Tested against `main` at `41b4896e90edc7ed03f8291d52d9acd95e2fff35`.

## Where

`prepare_rollback_resource` (the **fixed** resource path) carefully separates the three cases — `crates/replication/prediction/src/rollback.rs` (~1073–1104):

```rust
match history_value {
    // no record at-or-before rollback tick -> tracking hadn't started / sample
    // shifted past the target. Leave the live value in place.
    None => { /* leave current value */ }
    // authoritative record that it was absent -> remove.
    Some(HistoryState::Removed) => { commands.remove_resource::<R>(); }
    Some(HistoryState::Updated(r)) => { /* restore */ }
}
```

`prepare_rollback` (the **component** path) collapses "no record" and "removed" into a single `Option<C>` and treats both as a removal — `rollback.rs` (~966–1004):

```rust
let restore_value = if let Some(confirmed_restore) = confirmed_restore {
    confirmed_restore.into_value()          // ConfirmedState::Removed -> None, too
} else {
    predicted_history.get(rollback_tick).cloned()   // PredictionState::Removed AND "no entry" -> None
};
// ...
match restore_value {
    None => { entity_mut.try_remove::<C>(); }   // <-- "no sample" misread as "removed"
    Some(correct) => { /* insert / overwrite */ }
}
```

[...]

Issue #1487, opened 2026-05-28
State rollback churn: unchanged-entity rollback can be followed by confirmed-update rollback for same/adjacent window


rherv:
## Summary

I am seeing intermittent prolonged visual jitter on predicted entities in my 10 TPS game, roughly every couple of minutes. When I added rollback diagnostics, the suspicious pattern was that state rollback appeared to be requested/executed more than once around the same server-confirmed window.

After digging into it, I think there may be rollback churn between two state rollback request paths:

1. **unchanged-entity path**  
   `ServerMutateTicks` advances, but a predicted entity's `ConfirmHistory` is stale. The rollback checker infers the entity was unchanged at the newer server-confirmed tick and compares the last confirmed value against prediction history.

2. **confirmed-update path**  
   A later replicated component update is written into `PredictionHistory` through `PredictionRegistry::write_history`. If it mismatches prediction history, it records `StateRollbackMetadata::earliest_mismatch_tick`, which is then consumed by `check_rollback`.

The concerning shape is:

```text
unchanged_entity rollback
-> rollback/replay
-> confirmed_update mismatch for same or adjacent server window
-> second rollback/replay
```

In my live game logs, I observed the tighter version of this pattern: an `unchanged_entity` rollback and then a `confirmed_update` rollback for the same rollback window at the same local tick. That correlates with visible prediction jitter.

## Reproduction branch

I added diagnostic tests here:

<https://github.com/rherv/lightyear/tree/repro/rollback-churn-unchanged-confirmed>
The branch is not intended as a fix. It is meant to document the behavior and separate three things:

1. the minimal request lifecycle,
2. the exact same-tick/same-window churn shape,
3. the real Replicon/write-history pipeline that can produce the follow-up confirmed-update rollback.

The model tests are small and only show that the request lifecycle permits `unchanged_entity -> confirmed_update` follow-up behavior. The ignored diagnostic test reproduces the exact same-local-tick/same-window shape I saw in my logs, but it manually injects phase 2 with `trigger_rollback_check()`.

The strongest test is: `test_state_rollback_real_replication_can_chain_unchanged_then_confirmed_update`
This one uses real server-to-client replication. It shows:
```text
server mutates helper entity
-> ServerMutateTicks advances through real replication
-> predicted entity is checked by unchanged-entity path
-> rollback occurs

then server mutates the predicted entity itself
-> client receives update through normal write_history path
-> confirmed value mismatches PredictionHistory
[...]

PR #1618, opened 2026-07-28
fix(prediction): preserve pre-target history so deeper rollbacks restore canonically

### LINKED ISSUE #1617: `prepare_rollback` wipes canonical pre-target history; a deeper rollback in the same window replays from the current-tick live value
**Title:** `prepare_rollback` wipes canonical pre-target history; a deeper rollback in the same window replays from the current-tick live value

## Summary

`prepare_rollback` (lightyear_prediction `rollback.rs`) clears each component's whole `PredictionHistory<C>` on every rollback and re-anchors it at the rollback tick. When two rollbacks happen in the same frame window and the second one targets an OLDER tick than the first, its floor lookup finds no sample at-or-before its target (the first rollback destroyed those entries), so the history is seeded with the **current-tick live component value** — a future value injected into the past. The replay from that tick diverges permanently: for never-changing components the seeded value can additionally carry frame-interpolation/visual-correction state rather than simulation state; for changing ones it is simply the wrong tick's state. Reproduced end-to-end in a 3-client deterministic-replication setup at 100–150 ms simulated latency: after the first rollback (target T1), a second rollback (target T0 < T1) in the same window restores T0 from the live value at the current tick (~T1 + k), and every subsequent checksum mismatches until process end; runs without the churn stay bit-exact.

## Mechanism

```rust
predicted_history.clear();
if let Some(state) = restore_state.clone() {
    predicted_history.add_state(rollback_tick, state);
} else if let Some(current) = predicted_component.as_deref() {
    // seed with the current component value (#1602)
    predicted_history.add_state(rollback_tick, HistoryState::Updated(current.clone()));
}
```

`HistoryBuffer::get_state` is a floor lookup (newest entry ≤ tick). After the first rollback the buffer holds only entries ≥ T1 (the reseeds and the replayed records), so a second rollback to T0 < T1 gets `None` and lands in the live-seed arm — even though the client held perfectly canonical per-tick samples for T0 one frame earlier. The entries strictly newer than a rollback target are exactly what the replay rewrites; everything at-or-before the target is still the client's canonical past and is what a later, deeper rollback needs as its restore point. Clearing the whole buffer conflates the two.

This is the same "restoration isn't safe for sparse history" family as #1511 (the `Some(stale)` face), but on the destructive side: the bug is not that the floor sample is stale, it's that the floor sample was thrown away. Rollback churn in the wild is documented in #1487 (unchanged-entity and confirmed-update paths firing for adjacent windows); with churn, this code path fires reliably around deep correction windows such as a peer joining mid-game.

## Suggested fix

Discard only the invalidated window instead of the whole buffer:

```rust
predicted_history.clear_after_tick(rollback_tick);
```

Entries at-or-before the target are preserved (the replay never rewrites them), entries strictly newer are discarded and re-recorded by the replay. The `add_state` calls below the clear are unchanged: `add_state` replaces a same-tick entry, so re-anchoring at the target stays correct. Compile+test verified against current `main` (00083ff8 — cherry-picked cleanly onto it, 9/9 `lightyear_prediction` rollback tests pass) plus the following regression test, which fails before the change and passes after:

```rust
#[test]
fn test_rollback_preserves_pre_target_history_for_deeper_rollback() {
    let mut world = World::new();
    world.init_resource::<LocalTimeline>();
    world.init_resource::<PredictionRegistry>();
    world.init_resource::<ReplicationCheckpointMap>();

    let manager = world
        .spawn((PredictionManager::default(), Rollback::FromInputs))
        .id();

    let mut history = PredictionHistory::<TestComponent>::default();
    for tick in [8, 10, 12, 15] {
        history.add_predicted(Tick(tick), Some(TestComponent(tick as f32)));
    }
    let predicted = world.spawn((Predicted, TestComponent(15.0), history)).id();

    // First rollback to tick 12.
    world.get_mut::<PredictionManager>(manager).unwrap().set_rollback_tick(Tick(12));
    world.run_system_once(prepare_rollback::<TestComponent>).unwrap();

    let history = world.get::<PredictionHistory<TestComponent>>(predicted).unwrap();
    assert_eq!(history.get_state(Tick(8)), Some(&HistoryState::Updated(TestComponent(8.0))));
    assert_eq!(history.get_state(Tick(15)), Some(&HistoryState::Updated(TestComponent(12.0))));
    assert_eq!(world.get::<TestComponent>(predicted), Some(&TestComponent(12.0)));

    // Second, deeper rollback to tick 8 restores the preserved sample.
    world.get_mut::<PredictionManager>(manager).unwrap().set_rollback_tick(Tick(8));
    world.run_system_once(prepare_rollback::<TestComponent>).unwrap();
    assert_eq!(world.get::<TestComponent>(predicted), Some(&TestComponent(8.0)));
}
```

## Related

- Predicted component wrongly removed on state rollback when no history sample exists at-or-before the rollback tick #1511 — same family, restore side.
- State rollback churn: unchanged-entity rollback can be followed by confirmed-update rollback for same/adjacent window #1487 — the trigger that makes this fire.
- Rollback wipes the prediction history of never-changing components #1602 — the seed arm this change feeds into; the seed itself stays correct for entities revealed after the rollback target.

BoysGameStudio:
Fixes #1617.

`prepare_rollback` cleared the whole `PredictionHistory` on every rollback and re-anchored it at the rollback tick. Under rollback churn (two rollbacks in one window, the second deeper — see #1487), the deeper rollback then found no sample at-or-before its target and fell through to seeding the history with the current-tick live component value — a future value injected into the past, so the replay diverged permanently (same sparse-restore family as #1511).

Discard only the invalidated window instead: entries at-or-before the rollback target are preserved (the replay never rewrites them), so a later, deeper rollback finds its exact per-tick floor sample.

## Test plan

- Adds a churn regression test (two rollbacks, the second deeper, asserting the preserved sample is the restore source): fails before the change, passes after.
- `cargo test -p lightyear_prediction rollback`: 9/9 pass on current `main`.
- End-to-end in a 3-client deterministic-replication setup at 100–150 ms simulated latency: before, ~50–100% of runs cascaded (1000+ contiguous checksum mismatches after a mid-game join); after, repeated runs stay within the documented join transient.

### DIFF
--- a/crates/replication/prediction/src/rollback.rs
+++ b/crates/replication/prediction/src/rollback.rs
@@ -939,7 +939,18 @@ pub(crate) fn prepare_rollback<C: Component<Mutability = Mutable> + Clone>(
         // For completed state rollbacks this is the completed mutate tick; for
         // forced state and input rollbacks it may be older than the latest
         // completed mutate tick.
-        predicted_history.clear();
+        //
+        // Discard only the entries the rollback invalidates (strictly newer
+        // than the target): the replay rewrites those from the restored state.
+        // Entries at-or-before the target are still the client's canonical
+        // past — the replay never touches them — and they are exactly what a
+        // later, DEEPER rollback in the same frame window needs as its floor
+        // sample. Clearing the whole buffer here would throw them away, so a
+        // second rollback to an older tick would find no sample at-or-before
+        // its target and fall through to seeding the history with the
+        // current-tick live value — injecting a future value into the past
+        // (permanent divergence, see the Some(stale) class in issue #1511).
+        predicted_history.clear_after_tick(rollback_tick);
         if let Some(state) = restore_state.clone() {
             predicted_history.add_state(rollback_tick, state);
         }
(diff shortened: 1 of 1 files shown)


cBournhonesque:
Yes in general doing a deeper rollback in the same window as another rollback is not valid, since the prediction history has been cleared.
I guess my question is: how do you get two rollbacks in the same window? (by window, do you mean frame)?
The rollback checker should use the deepest mismatch tick identified across all entities.
For updated entities, we check for mismatch when writing the update.
For non-updated entities, we use the LastConfirmedTick received from replicon.

cBournhonesque:
Actually i think it is possible to have a later rollback be deeper than previous rollbacks if you mix Input rollbacks and State rollbacks.
It is relatively hard to guarantee the invariant that later rollback will **never** be deeper than previous ones, and I am not even sure it is something we want. (e.g. we receive an input mismatch so we rollback from tick 10 with fixed inputs, but then we receive an older mismatched state at tick 8 which shows that were still wrong, so we need to rollback again from tick 8). This is the situation in the avian_2d example where we predict other players.
We enable input rollback because inputs are received more frequently and faster than state, so we can early rollbacks whenever we notice that our input prediction was wrong. However that doesn't mean that we know the correct state (if the game is not deterministic) so we might still do deeper state-based rollbacks.

Could you add pruning?
```
  let oldest_legal_tick = current_tick - max_rollback_ticks;
  predicted_history.clear_until_tick(oldest_legal_tick);
  predicted_history.clear_after_tick(rollback_tick);
```
otherwise the history might grow indefinitely.

I will open another PR to add pruning independently from rollback (otherwise if there are no rollbacks the history will grow too large)

PR #1621, opened 2026-07-30
fix(raw): tolerate clients despawned during server stop

bugsweeper:
## Summary

- silence the deferred `Disconnecting` insertion when a raw client link was already despawned
- add a regression test covering a transport despawn queued before raw-server stop commands are applied

## Why

A transport disconnect can queue a client-link despawn in the same observer cycle as `RawConnectionPlugin::on_stop`. With Bevy 0.19, the later deferred `insert(Disconnecting)` panics when applied to that already-despawned entity. The link is already fully disconnected in this case, so treating the insertion as best-effort makes shutdown idempotent.


cBournhonesque:
Agreed, all these lifecycles insertions/removals should have a 'try' prefix

PR #1631, opened 2026-08-03
fix(deps): Replace unmaintained bincode for STD

naomijub:
Library `bincode` is considered unmaintained with [RUSTSEC-2025-0141](https://rustsec.org/advisories/RUSTSEC-2025-0141.html)m while library `bitcode` is still maintained. 

Another reason for this PR is the [serialization benchmarks](https://github.com/djkoloski/rust_serialization_benchmark), that considers `bincode` a bit slower than `bitcode`, which has proven correct in my tests for the suggested change.

### DIFF
--- a/crates/transport/serde/src/registry.rs
+++ b/crates/transport/serde/src/registry.rs
@@ -167,12 +167,13 @@ fn default_context_deserialize<C, M>(
 }
 
 #[cfg(feature = "std")]
-/// Default serialize function using bincode
+/// Default serialize function using bitcode
 fn default_serialize<M: Serialize>(
     message: &M,
     buffer: &mut Writer,
 ) -> Result<(), SerializationError> {
-    let _ = bincode::serde::encode_into_std_write(message, buffer, bincode::config::standard())?;
+    let encoded = bitcode::serialize(message)?;
+    buffer.extend_from_slice(&encoded);
     Ok(())
 }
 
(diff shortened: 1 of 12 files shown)


crates/transport/serde/src/registry.rs
```diff
@@ -167,12 +167,13 @@ fn default_context_deserialize<C, M>(
...
 
 #[cfg(feature = "std")]
-/// Default serialize function using bincode
+/// Default serialize function using bitcode
 fn default_serialize<M: Serialize>(
     message: &M,
     buffer: &mut Writer,
 ) -> Result<(), SerializationError> {
-    let _ = bincode::serde::encode_into_std_write(message, buffer, bincode::config::standard())?;
+    let encoded = bitcode::serialize(message)?;
```
cBournhonesque, line 175: I believe bitcode serializes into a new vec, and I really would like to avoid allocating in the hotpath! https://github.com/SoftbearStudios/bitcode/issues/6

I think it's possible to use bitcode::Buffer to make ser/de relatively cheap, but it would require types to implement Encode/Decode instead of serde: https://github.com/SoftbearStudios/bitcode/issues/102

cBournhonesque: Switched to postcard instead: https://github.com/cBournhonesque/lightyear/pull/1632

PR #1688, opened 2026-08-15
Make `MAX_CLIENTS` configurable

tidely:
Currently there is a `MAX_CLIENTS` constant defined inside of `lightyear_net`, which limits the amount of concurrent client connnections. This seems to be a legacy leftover, and clients are no longer maintained with slots, but rather dynamically. We should likely make this option configurable. I've added the old limit as the default for backwards compatibility, but it could be make an optional instead?


tidely:
This change seemed simple enough that I skipped creating an issue for it first, I hope that's okay :)

cBournhonesque:
Thanks!

Issue #888, opened 2025-02-09
Improve prediction


cBournhonesque:
- check_rollback query could run in parallel
- should we simplify some logic by always adding PredictedHistory<C> for all ComponentSyncMode::FULL as soon as the predicted entity is spawned (even if C wasn't added on the entity?)
- type-erased

PR #1696, opened 2026-08-17
fix(prediction): deliver server inserts of never-predicted components to Predicted entities

### LINKED ISSUE #1692: Server-inserted Full-mode component never applies to the Predicted entity unless an unrelated rollback fires
A `PredictionMode::Full` component that the **server inserts mid-game** on an already-spawned, already-predicted entity is never applied to the client's Predicted entity — unless an unrelated state rollback happens to fire, which applies it as a side effect.

Found in 0.29.0; the relevant code in `crates/replication/prediction/src/registry.rs` is identical on current main (checked 2026-08-16).

## Symptom

In our game (client/server, WebTransport, native inputs), death is a server-inserted `Dead` marker registered `.replicate().predict()`. The client cannot predict it (damage is server-only), so it never inserts `Dead` locally.

- The insert reliably reaches **Interpolated** entities. ✅
- **Mutations** of other Full components on the same Predicted entity (e.g. `Health`) sync reliably. ✅
- The insert **never arrives on the Predicted entity** — at any latency, over any number of deaths. ❌

The failure is connection-quality-dependent in a perverse way: a player with a poor connection (late/jittery inputs → frequent mispredictions → frequent rollbacks) receives the component shortly after each insert, because every rollback restore applies whatever is parked in `ConfirmedHistory<C>`. Players with clean connections (host-adjacent LAN, localhost) never receive it at all. In a 3-player session, only the highest-ping player had a working death screen — which made this fun to diagnose.

## Mechanism

As far as I can tell from reading `registry.rs`:

1. The replicon receive path for Full components (`write_history`) parks incoming values in `ConfirmedHistory<C>` rather than writing the live component. The only path that applies a buffered value to the entity is a **state-rollback restore**.
2. `should_rollback_check` does treat `(confirmed = Some, predicted = None)` as a mismatch — but it's only reachable when a mismatch check actually runs for that component.
3. The completed-tick scan (`check_rollback_for_unchanged_component`) returns `false` immediately when the entity has **no `PredictionHistory<C>`** — and a component the client never predicted has no `PredictionHistory<C>`.

So the arrival of a never-predicted component has no path that can flag it: it can't trigger a rollback itself, and nothing else applies it. It sits in `ConfirmedHistory<C>` until some *other* component's misprediction triggers a rollback. (The `// TODO: ideally we would update the LastConfirmedTick at this point?` above `write_history` suggests the receive path's bookkeeping is known to be incomplete here.)

## Reproduction

Minimal setup:

1. Register a marker-ish component `.replicate().predict()` that only the server ever inserts/removes (client has no system predicting it).
2. Server inserts it mid-game on a client's predicted entity.
3. Keep the client's prediction clean — a stationary client (no movement input) works well, since without mispredictions there are no rollbacks.

Observed with an insert-logging system on the client (`Added<C>` over all entities, ~90s sessions, repeated): dozens of arrivals with `interpolated=true`, **zero** with `predicted=true`, across 5–7 server-confirmed inserts targeting the predicted entity — while mutations of a different Full component on the same entity arrive exactly on schedule. Adding a receive-side link conditioner (60ms) changes nothing, consistent with the mechanism (inputs still arrive on time → no mispredictions → no rollbacks).

## Workaround

Derive client-side state from a Full component's **value** instead of its presence (we key "am I dead" off replicated `Health <= 0`, which the server makes coincide with the `Dead` insert). Works, but only when a value-companion exists.

## Possible fix directions

- The design note in #888 — "always add `PredictionHistory<C>` for all Full-mode components as soon as the predicted entity is spawned (even if C wasn't added)" — would make the existing `(Some, None)` mismatch check reachable and close this hole.
- Alternatively: when `write_history` records a confirmed value for a component that has no `PredictionHistory<C>` on the entity, treat that as an immediate rollback trigger (or apply the insert directly, mirroring what already happens for Interpolated).

## Related

- #1507 introduced the `ConfirmedHistory`/`PredictionHistory` split this lives in (its notes flag "future confirmed updates" as an edge, but not inserts of never-predicted components).
- #1511 — same restore path, inverse symptom (present component wrongly removed).
- #888 — contains the design note above.

Environment: lightyear 0.29.0, bevy 0.19.1, avian3d 0.7, WebTransport transport, native inputs (`lag_compensation: true`), 64Hz fixed tick.

pauldemarco:
Fixes #1692.

## Problem

A Full-mode component that the server inserts mid-game on an already-predicted entity — one the client never predicts itself (a server-authoritative `Dead`-style marker) — never reaches the Predicted entity unless an unrelated rollback happens to fire:

- the receive-time mismatch check in `record_confirmed_and_maybe_check` is often skipped: insert messages ride a reliable channel and can resolve to a tick at/behind `StateRollbackMetadata::last_processed_tick`;
- the unchanged-entity scan (`check_rollback_for_unchanged_component`) returns early when there's no retained predicted sample;
- and `prepare_rollback` uses `&mut PredictionHistory<C>` as its restore membership marker, so even when a rollback *does* fire, the component is skipped.

Net effect: the confirmed value is parked in `ConfirmedHistory<C>` forever. Clients with unstable connections receive the component as a side effect of their frequent rollbacks; clients with clean prediction never receive it at all.

## Fix

When a confirmed **value** arrives for a component the entity has neither predicted (no `PredictionHistory<C>`) nor currently has (no live `C`), seed `PredictionHistory<C>` with the client's actual prediction — `HistoryState::Removed` at the confirmed tick. That makes the entire existing pipeline work unmodified:

- the unchanged-entity scan now has a retained predicted sample (absence) to compare against the confirmed value → `(Some, None)` mismatch → state rollback at the next completed mutate tick;
- `prepare_rollback` now includes the entity for `C` and restores from `ConfirmedHistory` (at-or-before lookup finds the insert) → component inserted on the Predicted entity.

This is effectively a lazy, receive-scoped version of the note in #888 ("always add `PredictionHistory<C>` for all Full-mode components as soon as the predicted entity is spawned, even if C wasn't added") — scoped to components that actually receive confirmed data, so there's no per-spawn cost for unused components.

## Testing

- New regression test `test_late_server_insert_of_unpredicted_component_reaches_predicted`: real client-server stepper, predicted entity spawned without `CompFull`, server inserts it mid-game, no client misprediction. **Fails on current main** (component never arrives), passes with the fix.
- Full `lightyear_tests` suite: 189 passed / 0 failed single-threaded with the fix (parallel runs show pre-existing flakiness on stock main as well — global metrics recorder / shared state).
- Verified in our game (client/server, WebTransport, 64Hz, `lag_compensation` inputs) with a 60ms receive-side conditioner and a deliberately never-mispredicting client: on main, 0/8 mid-game marker inserts reached the predicted entity across 8 deaths; with this fix, 8/8 arrived, and interpolated-entity delivery and mutation sync are unchanged.

### DIFF
--- a/crates/replication/prediction/src/registry.rs
+++ b/crates/replication/prediction/src/registry.rs
@@ -473,6 +473,25 @@ impl PredictionRegistry {
 
         let predicted_history = entity_mut.get::<PredictionHistory<C>>();
 
+        // A confirmed VALUE arriving for a component this entity has never
+        // predicted (no PredictionHistory<C>) and does not currently have
+        // (no live C) — e.g. the server inserted a marker like `Dead`
+        // mid-game that the client cannot predict — is otherwise
+        // undeliverable (#1692):
+        // - the receive-time mismatch check below is often skipped because
+        //   insert messages ride a reliable channel and can resolve to a tick
+        //   at or behind `StateRollbackMetadata::last_processed_tick`;
+        // - the unchanged-entity scan (`check_rollback_for_unchanged_component`)
+        //   returns early without a retained predicted sample;
+        // - and `prepare_rollback` uses PredictionHistory<C> as its membership
+        //   marker, so even an unrelated rollback's restore skips C.
+        // Seed the prediction history with what the client actually predicted
+        // — "absent" — so the normal mismatch → rollback → restore pipeline
+        // picks the component up at the next completed mutate tick.
+        let never_predicted_insert = confirmed_component.is_some()
+            && predicted_history.is_none()
+            && entity_mut.get::<C>().is_none();
+
         #[cfg(feature = "metrics")]
         if let Some(predicted_history) = predicted_history.as_ref() {
             self.prediction_map[&ComponentKind::of::<C>()]
@@ -546,6 +565,19 @@ impl PredictionRegistry {
         } else {
             let mut history = ConfirmedHistory::<C>::default();
             history.insert(confirmed_tick, confirmed_state);
+            entity_mut.insert(history);
+        }
+        if never_predicted_insert {
+            trace!(
+                target: "lightyear_debug::prediction",
+                kind = "seed_prediction_history_for_unpredicted_insert",
+                entity = ?entity,
+                component = ?name,
+                confirmed_tick = confirmed_tick.0,
+                "seeding PredictionHistory with predicted-absence for a confirmed insert of a never-predicted component"
+            );
+            let mut history = PredictionHistory::<C>::default();
+            history.add_state(confirmed_tick, HistoryState::Removed);
             entity_mut.insert(history);
         }
         should_rollback
--- a/crates/tests/src/client_server/prediction/spawn.rs
+++ b/crates/tests/src/client_server/prediction/spawn.rs
@@ -1,3 +1,4 @@
+use crate::protocol::CompFull;
 use crate::stepper::*;
 use bevy::ecs::hierarchy::ChildOf;
 use lightyear_connection::network_target::NetworkTarget;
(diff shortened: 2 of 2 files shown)


cBournhonesque:
Great find, thanks

PR #1710, opened 2026-08-21
fix(rollback): Don't consume deferred check on same tick

tidely:
PR #1708 fixes confirmed updates that arrive ahead of the client timeline, but there is still an equal-tick case where the deferred check can be consumed too early.

# Reasoning

During replication receive, `record_confirmed_and_maybe_check` defers the comparison when:

`confirmed_tick >= current_tick`

Later in the same `PreUpdate`, `check_rollback` processes the completed server frontier when:

`server_confirmed_tick <= current_tick`

If the authoritative tick equals the current local tick:

1. The receive path records the entity in `pending_entity_state_checks`.
2. `check_rollback` immediately drains that entry with `take_through`.
3. `check_rollback_for_unchanged_component` looks for the predicted value at that tick.
4. That value may not exist yet because prediction history is recorded later in `FixedPostUpdate`.
5. The comparison returns `false`, but the pending entry has already been removed.
6. When prediction history becomes available, the authoritative state is never checked again.

# Fix

I've simply fixed the statement to not consume on the same tick, and I added a regression test for this case which can also be inspected for better context.

### DIFF
--- a/crates/replication/prediction/src/registry.rs
+++ b/crates/replication/prediction/src/registry.rs
@@ -2107,6 +2107,92 @@ mod tests {
     }
 
     #[test]
+    fn current_tick_confirmation_waits_for_prediction_history() {
+        let (mut app, fns_id) = setup_prediction_diff_app();
+        app.add_plugins(bevy_time::TimePlugin);
+        app.init_resource::<InputTimelineConfig>();
+        let client = app.world_mut().spawn_empty().id();
+        app.world_mut().resource_mut::<NetworkingMetadata>().mode = NetworkTopology::Client(client);
+        let mut sync = LocalTimelineSync::default();
+        sync.set_synced(true);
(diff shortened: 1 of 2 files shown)


cBournhonesque:
I think actually the issue is that in `record_confirmed_and_maybe_check` we skip the rollback check if `confirmed_tick >= current_tick` but actually we should be doing it if confirmed_tick == current_tick.

When the local_tick is T at PreUpdate, that means that the **previous frame** ran tick `T` so the PredictionHistory has a value for that tick.

Thank you for your PR; I will open a separate one that fixes that. (and also that simplifies the tests by moving them to lightyear_tests where we have better setup

cBournhonesque:
Closed in favor of https://github.com/cBournhonesque/lightyear/pull/1711

PR #1751, opened 2026-09-15
fix(prediction): use try_remove when dropping VisualCorrection on component removal

### LINKED ISSUE #1750: `remove_correction_state_on_live_removed` warns on every despawn of a corrected entity
Since 0.30, despawning any entity that has a corrected component (`Position`/`Rotation` in our case) logs a warning for each component:

```
WARN bevy_ecs::error::handler: Encountered an error in command `<bevy_ecs::system::commands::entity_command::remove<lightyear_prediction::correction::VisualCorrection<avian3d::physics_transform::transform::Position>>::{{closure}} as bevy_ecs::system::commands::entity_command::EntityCommand>::with_entity::{{closure}}`: Entity despawned: The entity with ID 1141v1 is invalid; its index now has generation 2.
```

It comes from the observer added in `correction.rs`:

```rust
fn remove_correction_state_on_live_removed<C: Component, D: Send + Sync + 'static>(
    trigger: On<Remove, C>,
    mut commands: Commands,
) {
    commands
        .entity(trigger.entity)
        .remove::<VisualCorrection<D>>();
}
```

`On<Remove, C>` also fires when the whole entity is despawned. The entity is still alive when the observer runs, so the command gets queued fine, but by the time it's applied the entity is gone and Bevy reports the error.

The entity doesn't have to be predicted. In our case the first ones to hit it were plain static colliders (`RigidBody::Static`, which gets `Position`/`Rotation` from avian) that the client despawns when a gate opens. Projectiles and players at match end hit it the same way.

Nothing actually breaks, the component goes away with the entity anyway. It's just log noise on every despawn.

Suggested fix is `try_remove`:

```rust
commands
    .entity(trigger.entity)
    .try_remove::<VisualCorrection<D>>();
```

Versions: lightyear 0.30.0, bevy 0.19.1, avian3d 0.7.0. Didn't happen on 0.29, the observer isn't there.

lapsha38:
Fixes #1750. The On<Remove, C> observer also fires on despawn, so the queued remove runs against an entity that no longer exists and logs a warning. try_remove skips it.


cBournhonesque:
Good catch, thanks

